PR #38, opened 2021-10-19
Docker image publish

afonsojramos:
Publishes the docker image to the GitHub Container Repository in order to show up in the Packages section.

Check out a successful deploy [here](https://github.com/afonsojramos/parrot/runs/3945442582?check_suite_focus=true) and check out the published container [here](https://github.com/afonsojramos/parrot/pkgs/container/parrot).

This will later facilitate the installation for users without the need to build the image from scratch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,9 @@
 name: Main Workflow
-on: [push]
+
+on:
+  push:
+  pull_request:
+
 jobs:
   build:
     name: Build


aquelemiguel:
@afonsojramos How do GitHub packages diverge from something like Docker Hub?
I know very little about Docker, so I'm trusting your inputs here, @joao-conde @rafaeldamasceno.

The only real rule is that we cannot host the bot ourselves because Parrot (purposefully) uses `youtube-dl` (bad rep with Google). If this facilitates the deployment process for people wanting to use the bot, I'm all for it.

joao-conde:
LGTM, @afonsojramos can you find some formatter/prettier for YML files and run the ones we have?

PR #55, opened 2021-11-26
Refactor play

### LINKED ISSUE #43: Reduce `!play` duplicate code
Right now, there are some instances where code is needlessly duplicated:

- Before playing, the `!play` command summons the bot into a voice chat if it wasn't in one before.
The `!summon` command does the same.

- The `!play` and `!playtop` commands do exactly the same except for the position where the new track is inserted.
So, for instance, changing the behavior of adding playlists requires one to change both files. (This is very bad!)

- If a new track was enqueued via `!play` or `!playtop` and the queue was previously empty, a **Now Playing** message is sent.
However, the user can also get this message via `!nowplaying` and code is duplicated there.

- Maybe others I'm not seeing right now.

Solving this comes down to identifying the duplicate code and exporting it to its own functions to allow sharing between commands.

aquelemiguel:
* Closes #43.


aquelemiguel:
@afonsojramos Off topic, should we be building Docker images on every pull request?
Shouldn't it only be when `main` gets changed?

afonsojramos:
@aquelemiguel If you check out the builds of a PR you will notice that 3 workflows have run. One for the Docker and 2 for the build. This is because the Docker one, if I remember correctly, only runs on main and PR. What this means is that we are not checking on every push (like the build workflow), but once a PR is created we are checking if the Docker build is not breaking due to the changes within that PR so that you can confidently merge.

Furthermore, we are not publishing this Docker image, that is only once it gets merged if I'm not mistaken.

src/commands/play.rs
```diff
@@ -1,22 +1,156 @@
...
+
+                for entry in &entries {
+                    let uri = format!(
+                        "https://www.youtube.com/watch?v={}",
+                        entry.url.as_ref().unwrap()
+                    );
+                    let source = Restartable::ytdl(uri, true).await.unwrap();
+                    let mut handler = call.lock().await;
+                    handler.enqueue_source(source.into());
+                }
```
joao-conde, line 75: Can't you just use `enqueue_song` for each entry?

src/commands/play.rs
```diff
@@ -1,22 +1,156 @@
...
+                    let rotations = not_playing.len() - &entries.len();
+                    not_playing.rotate_left(rotations);
+                    queue.append(&mut not_playing);
+                });
+
+                queue_snapshot = handler.queue().current_queue();
+            }
+        }
+        _ => (),
+    }
```
joao-conde, line 101: Can't you just get all the `entries` and process them with `enqueue_song` with the `PlayFlag` passed?

src/commands/play.rs
```diff
@@ -1,22 +1,156 @@
...
+
+    queue_snapshot
+}
+
+async fn send_added_to_queue_message(
```
joao-conde, line 106: Shouldn't this be moved to the same place as `send_simple_message`? Either utils or a new module for message notifications or something.

src/commands/play.rs
```diff
@@ -32,133 +166,62 @@ async fn play(ctx: &Context, msg: &Message, mut args: Args) -> CommandResult {
...
+    let call = manager.get(guild.id).unwrap();
+    let enqueue_type: EnqueueType;
+
+    if url.clone().contains("youtube.com/playlist?list=") {
+        enqueue_type = EnqueueType::PLAYLIST;
+    } else if url.clone().starts_with("http") {
+        enqueue_type = EnqueueType::URI;
+    } else {
+        enqueue_type = EnqueueType::SEARCH;
+    }
```
joao-conde, line 186: Nope. Do:

```rust
let enqueue_type = if url.clone().contains("youtube.com/playlist?list=") {
        EnqueueType::PLAYLIST
    } else if url.clone().starts_with("http") {
        EnqueueType::URI
    } else {
        EnqueueType::SEARCH
    }
```

joao-conde (changes requested):

joao-conde:
@aquelemiguel assign me and re-request review when you are done

src/commands/play.rs
```diff
@@ -1,22 +1,156 @@
...
+    };
+
+    let mut handler = call.lock().await;
+    handler.enqueue_source(source.into());
+    let mut queue_snapshot = handler.queue().current_queue();
+    drop(handler);
+
+    match flag {
+        PlayFlag::PLAYTOP => {
+            if queue_snapshot.len() > 2 {
```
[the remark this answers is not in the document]
aquelemiguel: Your rationale is correct, but at this point the track has already been inserted:
* If there are now two songs in the queue (one playing and one waiting), there's no need to do anything, the next track is already at the top.
* If there are now three songs, take the last one and rotate it to the second position (under the playing one).

src/commands/play.rs
aquelemiguel: Good thinking but I don't think so - at least with how the code is structured right now. Because `enqueue_song` places the song at the top of the queue, the playlist should be enqueued in reverse. I could reverse the list of songs in a playlist and pass it to `enqueue_song`. A little hacky but should eliminate the last match in this method, no?

src/commands/play.rs
aquelemiguel: [See discussion above.](https://github.com/aquelemiguel/parrot/pull/55#discussion_r775577954)

src/commands/play.rs
aquelemiguel: Agreed, moved.

src/commands/play.rs
aquelemiguel: Beautiful, changed.

src/commands/play.rs
```diff
@@ -1,22 +1,156 @@
...
+async fn enqueue_song(
+    call: &Arc<Mutex<Call>>,
+    query: String,
+    is_url: bool,
+    flag: &PlayFlag,
+) -> Vec<TrackHandle> {
+    let source = match is_url {
+        true => Restartable::ytdl(query, true).await.unwrap(),
+        false => Restartable::ytdl_search(query, false).await.unwrap(),
+    };
```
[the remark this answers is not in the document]
joao-conde: Yo

src/commands/play.rs
joao-conde: I think it makes sense to reuse `enqueue_song`. Calling reverse on the list of songs to enqueue is good enough. Alternatively, that can be a boolean on the enqueue song method like "reverse"

joao-conde (changes requested):

src/commands/play.rs
aquelemiguel: Well, I didn't think this through because imagine playlist `[A,B,C,D]` and an empty queue.
If you `!playtop` the reversed playlist, you get `[D,A,B,C]` - D immediately starts playing and the rest get pushed under.

Do you think it is that useful to pushtop a playlist? I'm currently considering removing it to simplify the code.

src/commands/play.rs
```diff
@@ -1,22 +1,78 @@
...
     client::Context,
     framework::standard::{macros::command, Args, CommandResult},
     model::channel::Message,
+    prelude::Mutex,
 };
 
-use songbird::input::Restartable;
-
+use songbird::{input::Restartable, Call};
 use youtube_dl::{YoutubeDl, YoutubeDlOutput};
```
joao-conde, line 18: I would reorder functions as play -> execute_play -> enqueue_playlist -> enqueue_song

src/commands/play.rs
```diff
@@ -1,22 +1,78 @@
...
+                queue.append(&mut not_playing);
+            });
+        }
+    }
+}
+
+async fn enqueue_playlist(call: &Arc<Mutex<Call>>, uri: &str) {
+    match YoutubeDl::new(uri).flat_playlist(true).run() {
+        Ok(result) => {
+            if let YoutubeDlOutput::Playlist(playlist) = result {
```
joao-conde, line 54: Not sure, but isn't this possible:

```rust
match ... {
   Ok(YoutubeDlOutput::Playlist(playlist)) =>
...
```

Or something of the sorts

src/commands/play.rs
joao-conde: Can't you:

1- `[A, B, C, D]` => play(A) + enqueue_playlist([D, C, B])

or

2- `[A, B, C, D]` => enqueue_playlist([A, B, C, D], reverse=false)

joao-conde (approved):

Issue #47, opened 2021-10-22
Restrict queue modifying commands to a DJ role


aquelemiguel:
Commands that perform modifications to the queue or make the bot connect/disconnect should be locked to a role or be put behind a voting system (relates to #44). As of writing, this is every command except for informative ones, which should stay unmodified:
- `now_playing`
- `queue`
- ~`genius/explain`~
- ~`genius/lyrics`~

If a user does not have said role and tries to use the command, a message informing their lack of permissions should be sent to the channel. A good discussion point would be how to tell the bot which role is the DJ role (relates to #45's discussion about where to store the prefix).

Antonio-Bennett:
Hey @aquelemiguel :) for this implementation how would you go about adding the role to a user. I know serenity has the User type and it can have flags but there isn't a DJ flag or anything. There is a moderator flag etc, however. https://docs.rs/serenity/0.10.9/serenity/model/user/struct.UserPublicFlags.html What were your thoughts on this?

Antonio-Bennett:
Hey @aquelemiguel @joao-conde I made this pull request that enables this functionality

joao-conde:
Hey @Antonio-Bennett !  We will be reviewing this ASAP and requesting changes or approving it right away. Thanks!

Antonio-Bennett:
Hey @joao-conde :) Thanks for the quick reply! For sure let me know if anything needs to change it was fun figuring this one out :)

aquelemiguel:
Don't know how to use it but this looks like the new way to [set permissions for application commands](https://docs.rs/serenity/latest/serenity/builder/struct.CreateApplicationCommandsPermissions.html#). Could be useful.

joao-conde:
Yup, I am going to use that, I discussed it in Serenity's discord

PR #56, opened 2021-12-14
Only able to use queue modifying commands if you have the DJ role

Antonio-Bennett:
| - | - |
| --- | --- |
| Issue | https://github.com/aquelemiguel/parrot/issues/47 |
| Dependencies | |
| Decisions | Added a new utility function that checks if the user has the DJ role. Also added a new string to use with send_simple_message utility. I did not create an add_role and delete_role command because that is already built into discord server settings really well. A moderator can just click the user's name and add the role. I can add it if requested :) |


aquelemiguel:
Hi @Antonio-Bennett, always good seeing you back here. 😊

I quickly glanced at your code, but when writing #47 I thought about locking the commands via a macro.
Please take a look at [`#[allowed_roles(roles)]`](https://docs.rs/serenity/0.10.9/serenity/framework/standard/macros/attr.command.html) and see whether this simplifies your code. 🙂

Antonio-Bennett:
@aquelemiguel Hey I love working on Parrot :)  Yeah, wouldn't that literally mean I just have to add that Macro to the commands.... Wow lol did not see that in the docs. Just 1 line 😭

joao-conde:
As @aquelemiguel  said, @Antonio-Bennett swap out for the macro usage and then tag us again for review! Thank you!

Antonio-Bennett:
@joao-conde Hey yeah I will do tomorrow sorry for the MIA yesterday was the last day of school so final projects and stuff :) I'll swap out later today and tag for review 🙂

Antonio-Bennett:
Hey @joao-conde @aquelemiguel I might be misinterpreting the docs for allowed roles I add this macro after the command macro but doesn't seem to do anything for me
image

aquelemiguel:
@Antonio-Bennett Huh, that's odd, your usage looks correct. I'm assuming you didn't forget to give yourself the "DJ" role, right?
I'll try it later today on my side and report back. 🤔

Antonio-Bennett:
@aquelemiguel The problem is it doesn't stop a user from using the command without the Role. So even with the macro, I can use the `!play` command with or without the `DJ` role. It's weird for sure.

aquelemiguel:
@Antonio-Bennett Interesting, do you also have the Administrator role by any chance? Because it might be overlapping with the DJ role?

Antonio-Bennett:
Hey @aquelemiguel I'm testing on a server that I made. Does the owner have an Administrator role by default?

aquelemiguel:
@Antonio-Bennett Looks like it!

image

Antonio-Bennett:
@aquelemiguel So what step do you want to take regarding this? I don't think it's possible to take away that role from owners. Also I guess your testing worked fine with the macro?

aquelemiguel:
> @aquelemiguel So what step do you want to take regarding this? I don't think it's possible to take away that role from owners. Also I guess your testing worked fine with the macro?

Just tested it with @joao-conde, yeah looks like we're right! (Ignore the orange parrot, that's our deployed instance)
I locked `!stop` behind the DJ role, like you did. I am administrator so my commands always went through. When giving him the role, he was able to use it again.

image

So I'd say you can freely add the macro to the commands you see fit. 😊

Antonio-Bennett:
Ok great will add it to the rest of commands now then :)

Antonio-Bennett:
@aquelemiguel wait I actually have one last question 😅 how exactly would we display the message that informs the user that they lack permission as stated in the issue? The macro kind of just blocks the use right? Is there a way to catch if it blocked it?

joao-conde:
@Antonio-Bennett we have no idea as well, that is a matter of searching a bit. Try to find some resources online and bring them for discussion here.

joao-conde (changes requested):
Use the macro approach to block, we don't really care about a message informing the blocked user. However, search a bit and if you find an easy enough solution to do so, apply it.

joao-conde:
@Antonio-Bennett I did a quick search and found this:

- https://docs.rs/serenity/0.10.9/serenity/framework/standard/enum.DispatchError.html#variant.LackingRole
- https://github.com/serenity-rs/serenity/blob/c3358cf6b251d9c3e6b4bfa8aba7b953b33074a9/examples/e05_command_framework/src/main.rs#L189

Please read it and follow a similar approach.

Also, new PRs entered master so make sure to fix conflicts and then re-assign to me or any other maintainer.

joao-conde:
Closing this due to inactivity and taking the issue for myself.

PR #58, opened 2021-12-29
Fix incorrect estimated time until play

### LINKED ISSUE #46: Incorrect estimated time until play
The estimated time until play is being calculated incorrectly - it should be the sum of the all the non-elapsed song durations before.
For instance, if song A is currently being played and is at `4:00 / 30:00` and song B is next and is `2:00` minutes long, then song C's estimated time until play should be `30:00 - 4:00 + 2:00 = 28:00`.

### Example

image

aquelemiguel:
* Closes #46

### DIFF
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use crate::{
     commands::{now_playing::now_playing, summon::summon, EnqueueType, PlayFlag},
@@ -13,7 +13,7 @@ use serenity::{
     prelude::Mutex,
 };
 
-use songbird::{input::Restartable, Call};
+use songbird::{input::Restartable, tracks::TrackHandle, Call};
 use youtube_dl::{YoutubeDl, YoutubeDlOutput};
 
 #[command]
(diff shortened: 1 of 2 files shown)


src/commands/play.rs
```diff
@@ -102,6 +118,42 @@ pub async fn execute_play(
...
+            .expect("Could not get playing track info")
+            .position;
+
+        let top_track_duration = top_track
+            .metadata()
+            .duration
+            .expect("Could not fetch duration of top track");
+
+        let mut estimated_time = match flag {
+            PlayFlag::DEFAULT => queue[1..queue.len() - 1]
```
joao-conde, line 137: Can just do `queue[1..]`

src/commands/play.rs
```diff
@@ -102,6 +118,42 @@ pub async fn execute_play(
...
+                    acc + x
+                        .metadata()
+                        .duration
+                        .expect("Could not fetch duration of track")
+                }),
+            PlayFlag::PLAYTOP => Duration::ZERO,
+        };
+
+        // Add the remaining top track
+        estimated_time += top_track_duration - top_track_position;
```
joao-conde, line 149: So `top_track_position` is not like a position index (0 1 2 or 3) right? I think its something else and perhaps rename it.

src/commands/play.rs
```diff
@@ -102,6 +118,42 @@ pub async fn execute_play(
     Ok(())
 }
 
+async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Option<Duration> {
```
joao-conde, line 121: Very nice function, will become even better when we add the error propagation mechanisms and just unwrap values with `?`

joao-conde (changes requested):

src/commands/play.rs
aquelemiguel: Nope, I don't want the last track, as it is the last inserted track.

src/commands/play.rs
aquelemiguel: 100% 🚀

src/commands/play.rs
aquelemiguel: Good point. In this context, position is the elapsed time for a track (that's what Songbird calls it), but it does make you think it is a index. I'll rename it to something more clear.

src/commands/play.rs
joao-conde: :donkey: me

joao-conde (approved):

PR #61, opened 2021-12-30
Replace `youtube-dl` dependency with `yt-dlp`

### LINKED ISSUE #59: Large seek freezing
## Description

The `!seek` command is blocking the bot. The larger the seek, the more time Parrot spends processing the request.

Seeking to further points in the track always yields larger and larger block times and is not relative to the current track position. This means that even if, for instance, the user successfully seeks to `4:00`, seeking to `4:05` takes a little bit longer than the initial one.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | Seeking a track is fast |
| **Observed** | Seeking a track makes the bot irresponsive for way too long |

## Repro Steps

1. Play any moderately lengthy track.
2. Try to seek to a timestamp of about 10 seconds. The bot should stop playing for a while but resume activity after a few seconds.
3. Now seek to a larger timestamp of a couple of minutes. Now the bot should be irresponsive for a substantially larger amount of time.

## Environment

| Key | Value |
| --- | --- |
| **Device** | N/A |
| **Operating System** | Windows 11 Enterprise |

## Screens

N/A

aquelemiguel:
Closes #59. Turns out #59 was happening because of `youtube_dl`. Replacing it with `yt-dlp` made seeking super fast.

Now given that Parrot no longer uses `youtube_dl`, we must update both our Dockerfile and README. Could any of you - @afonsojramos or @rafaeldamasceno - have a look at this? To reiterate, we need [`yt-dlp`](https://github.com/yt-dlp/yt-dlp) now.

Also I fucked up and committed these changes directly to master a while ago, I've gone ahead and deleted the bad image.


Cargo.toml
```diff
@@ -18,8 +18,9 @@ scraper = "0.12.0"
 regex = "1.5.4"
 
 [dependencies.songbird]
-version = "0.2.0"
-features = ["builtin-queue"]
+git = "https://github.com/serenity-rs/songbird"
+branch = "current"
```
joao-conde, line 22: Still, test it with songbird 2.0 maybe, it is not great to be using bleeding edge master version.

README.md
```diff
@@ -30,9 +33,9 @@ apt install build-essential autoconf automake libtool m4 libopus-dev youtube-dl
 ### Windows
 
 If you are using the MSVC toolchain, a prebuilt DLL for Opus is already provided for you.  
-You will only need to download [FFmpeg](https://ffmpeg.org/download.html), and install youtube-dl which can be done through Python's package manager, pip.
+You will only need to download [FFmpeg](https://ffmpeg.org/download.html), and install [yt-dlp](https://github.com/yt-dlp/yt-dlp) which can be done through Python's package manager, pip.
 ```shell
-pip install youtube_dl
+pip install -U yt-dlp
 ```
```
joao-conde, line 39: I would even go ahead and delete this, as the ffmpeg and yt-dlp readmes already tell you how to install

joao-conde (changes requested):

Cargo.toml
aquelemiguel: Tried it, doesn't compile because it has no support for a "yt-dlp" feature. 😐

joao-conde:
@aquelemiguel I have added it to the Dockerfile.

README.md
aquelemiguel: I would be in favour of simplifying the README and delegate install instructions to the respective repos. Maybe we do it at a later time?

joao-conde (approved):

PR #69, opened 2022-01-06
Allow prefix change

### LINKED ISSUE #45: Allow prefix change
Forcing users to use the `!` prefix can cause conflicts with other bots on the server.

Should this be an environment variable (easier but requires redeploying) or does it justify a whole new `!prefix` command usable only by admins (harder but allows for changing on the fly)? Requesting your inputs here, @afonsojramos and @joao-conde.

Regardless of the chosen approach:
* The bot's activity should also reflect the prefix change.
* Existing messages that reference the exclamation mark prefix should now use the current prefix.

afonsojramos:
* Closes #45 

Implementation details are rather straight up, however, for the HashMap merge with the `serde_json::Value`, in order to not do a single insert I used a feature that was implemented in Rust version `1.56` as it can be seen [here](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-From%3C%5B(K%2C%20V)%3B%20N%5D%3E).

Example interaction with bot:

image


src/commands/prefix.rs
```diff
@@ -0,0 +1,47 @@
...
+            HashMap::from([(guild_id.0.to_string(), new_prefix)]);
+        let new_prefixes = merge(&prefixes, &guild_new_prefix).await;
+        fs::write("prefixes.json", new_prefixes.to_string()).unwrap();
+
+        return send_simple_message(&ctx.http, msg, PREFIX_UPDATED).await;
+    } else if true {
+        return send_simple_message(&ctx.http, msg, MISSING_PREFIX).await;
+    }
+
+    Ok(())
```
joao-conde, line 33: You can remove this last `Ok(())` and remove the `return`s from the `if/else` above

src/commands/prefix.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+async fn merge(v: &Value, fields: &HashMap<String, String>) -> Value {
```
joao-conde, line 36: no need to be async, there is no await

joao-conde, line 36: I think this should merge a JSON `Value` with another `Value`.

Right now it only merges at the top level of the JSON and requires a hashmap. Try to look into this:
https://github.com/serde-rs/json/issues/377#issuecomment-341490464

src/commands/prefix.rs
```diff
@@ -0,0 +1,47 @@
...
+    let guild_id = msg.guild(&ctx.cache).await.unwrap().id;
+
+    let prefixes = get_prefixes();
+    let new_prefix = String::from(args.rewind().rest());
+
+    if !new_prefix.is_empty() && new_prefix.len() == 1 {
+        let guild_new_prefix: HashMap<String, String> =
+            HashMap::from([(guild_id.0.to_string(), new_prefix)]);
+        let new_prefixes = merge(&prefixes, &guild_new_prefix).await;
+        fs::write("prefixes.json", new_prefixes.to_string()).unwrap();
```
joao-conde, line 26: Here instead of unwrapping we could also send a message, saying "failed to save" or something.

src/main.rs
```diff
@@ -72,7 +76,26 @@ async fn main() {
...
+            c.dynamic_prefix(|ctx, msg| {
+                Box::pin(async move {
+                    let prefixes = get_prefixes();
+
+                    let guild_id = msg.guild(&ctx.cache).await.unwrap().id;
+
+                    if let Some(serde_json::Value::String(prefix)) =
+                        prefixes.get(guild_id.0.to_string())
+                    {
+                        // Command from guild_id.0 with prefix
```
joao-conde, line 89: These comments are superfluous

src/main.rs
```diff
@@ -72,7 +76,26 @@ async fn main() {
...
+                        Some(prefix.clone())
+                    } else {
+                        // Command from guild_id.0 with default prefix
```
joao-conde, line 92: These comments are superfluous

src/strings.rs
```diff
@@ -20,4 +20,9 @@ pub const MISSING_PLAY_QUERY: &str =
     "Missing query for this command! Either add a URL or keywords.";
 
 pub const MISSING_INDEX_QUEUE: &str = "Missing an index!";
+
 pub const NO_SONG_ON_INDEX: &str = "There is no queued song on that index!";
+
+pub const PREFIX_UPDATED: &str = "Updated prefix!";
+
+pub const MISSING_PREFIX: &str = "⚠️ Please send a character as an argument.\nExample: `!prefix ~`";
```
joao-conde, line 28: Nice unicode emoji. @aquelemiguel  would be proud.

src/utils.rs
```diff
@@ -64,3 +69,20 @@ pub fn get_human_readable_timestamp(duration: Duration) -> String {
 pub fn get_full_username(user: &User) -> String {
     format!("{}#{:04}", user.name, user.discriminator)
 }
+
+pub fn get_prefixes() -> serde_json::Value {
+    let data = fs::read_to_string("prefixes.json").unwrap_or_else(|error| {
+        if error.kind() == ErrorKind::NotFound {
+            let mut f = File::create("prefixes.json").unwrap_or_else(|error| {
+                panic!("Problem creating the file: {:?}", error);
+            });
```
joao-conde, line 78: This is the same as 

```rust
let mut f = File::create("prefixes.json").unwrap("Problem creating the file: {:?}", error);
```

src/utils.rs
```diff
@@ -64,3 +69,20 @@ pub fn get_human_readable_timestamp(duration: Duration) -> String {
...
+            f.write_all("{}".as_bytes())
+                .expect("Unable to write to file");
+            "{}".to_string()
```
joao-conde, line 81: Why this? It is not returned

joao-conde (changes requested):
I did not test it locally yet. I will do it in the evening to test functionality. But I commented some code changes.

src/commands/prefix.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        return send_simple_message(&ctx.http, msg, PREFIX_UPDATED).await;
+    } else if true {
```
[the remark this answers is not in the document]
afonsojramos: It was 4am :)

.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   build:
     strategy:
       matrix:
-        rust-version: ["1.54", "1.55", "1.56", "1.57"]
+        rust-version: ["1.56", "1.57"]
```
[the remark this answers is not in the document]
afonsojramos: Quoting from the PR:

> Implementation details are rather straight up, however, for the HashMap merge with the `serde_json::Value`, in order to not do a single insert I used a feature that was implemented in Rust version `1.56` as it can be seen [here](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-From%3C%5B(K%2C%20V)%3B%20N%5D%3E).

src/main.rs
afonsojramos: I thought so, but chose to leave them at the time, will remove.

src/strings.rs
afonsojramos: I think we should have more actually :eyes:

src/utils.rs
afonsojramos: We need to initialize the file as a valid json, otherwise it may fail later on.

src/commands/prefix.rs
afonsojramos: As you prefer my dear sir.

src/utils.rs
joao-conde: This: https://github.com/afonsojramos/parrot/blob/afonso/prefix-change/src/utils.rs#L81
is not returning it.

This is: https://github.com/afonsojramos/parrot/blob/afonso/prefix-change/src/utils.rs#L87

afonsojramos: Actually, it is returning the string to the `data` variable.

src/utils.rs
```diff
@@ -64,3 +71,17 @@ pub fn get_human_readable_timestamp(duration: Duration) -> String {
...
+pub fn get_prefixes() -> serde_json::Value {
+    let data = fs::read_to_string("prefixes.json").unwrap_or_else(|error| {
+        if error.kind() == ErrorKind::NotFound {
+            let mut f = File::create("prefixes.json").expect(FAILED_CREATE_PREFIXES);
+            f.write_all("{}".as_bytes()).expect(FAILED_SAVE_PREFIXES);
+            "{}".to_string()
+        } else {
+            panic!("Problem opening the file: {:?}", error);
+        }
+    });
```
joao-conde, line 84: Take a look into https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#examples

joao-conde (changes requested):

src/utils.rs
afonsojramos: Sure, but since in this case I need the file initialised with something, it is more useful with the current implementation.

joao-conde: I have no time now, but please let's hold this until tonight as I might have an idea. If that does not work, this is not by all means bad or wrong, so in that case we consider this resolved.

afonsojramos: What are we doing tonight? 😏

src/utils.rs
```diff
@@ -64,3 +70,30 @@ pub fn get_human_readable_timestamp(duration: Duration) -> String {
 pub fn get_full_username(user: &User) -> String {
     format!("{}#{:04}", user.name, user.discriminator)
 }
+
+pub fn get_prefixes() -> serde_json::Value {
```
joao-conde, line 74: I suggest [this](https://github.com/aquelemiguel/parrot/blob/jc/prefix-change-proposal/src/utils.rs#L74)

joao-conde (changes requested):
Also, let's add a `PREFIX_FILE` constant to `strings.rs` containing the path to the file. For now, I would keep it the same (`"prefixes.json"`)

aquelemiguel (changes requested):
The prefix command should be locked via [`#[required_permissions(perms)]`](https://docs.rs/serenity/latest/serenity/framework/standard/macros/attr.command.html). Probably keep it administrator only. Don't bother about informing non-admins they lack permission to use the command, leave it for #47.

joao-conde (approved):

aquelemiguel (approved):

PR #74, opened 2022-01-08
fix: bot activity using configured prefix

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Bot prefix does not have to be the default one. Shows correct activity now.<br>Refactor magic constant being used as the default prefix and replace its usages. |

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -1,3 +1,5 @@
+pub const DEFAULT_PREFIX: &str = "!";
+
 pub const IDLE_ALERT: &str =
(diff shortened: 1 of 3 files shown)


src/strings.rs
```diff
@@ -1,3 +1,5 @@
+pub const DEFAULT_PREFIX: &str = "!";
```
afonsojramos, line 1: I would say that we should change this to what is set in `.env.example` aka `-`.

afonsojramos (approved):

PR #83, opened 2022-01-10
version: 0.4.1

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Bugfixing only, increase `patch` |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parrot"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["aquelemiguel"]
 edition = "2018"
 description = "A Discord music bot built in Rust"


joao-conde:
@afonsojramos  @aquelemiguel  Let's follow this way for new versions, with a PR named `version: major.minor.patch`, a brief decision on what was increased, update `Cargo.toml` and tag this PR as `release`. Finally, once this is merged, do a new tag release.

afonsojramos (changes requested):
This has already been used during testing. The `0.4.X` should be marked as a dead release.

afonsojramos:
Additionally, even though this was a bugfix, you fundamentally changed the way something works, as such, it should not be considered a patch.

joao-conde:
@afonsojramos it is not a new feature for the crate users though.

afonsojramos:
@joao-conde At the moment, versioning is mostly only affecting docker users as we are not publishing the crate anywhere. As such, if `v0.4.x` never worked for docker users, you nuke that minor.

afonsojramos:
In the end, here we are managing the version of two distinct things: The cargo crate and the docker container, and these should be in sync. 

In this case there are "major" (not to confuse with major release) changes to the docker container, as such, we should move on to `v0.5.0`.

PR #95, opened 2022-01-21
remove command UX enhancement

### LINKED ISSUE #93: Improve remove index's tooltip and parameter type
## Rationale
Currently, when selecting a track to be removed from the queue, it isn't immediately obvious how it is going to work.

## Description

The UX of selecting the track to be removed could be improved with a tooltip that better clarifies how the queue indexing works. Additionally, the index parameter could be converted to an `INTEGER` option.

## Reference
[Application Command Option Type](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-type)

afonsojramos:
Closes #93.

* Changed the tooltip:
image

* Added handling support for negative numbers

### DIFF
--- a/src/commands/remove.rs
+++ b/src/commands/remove.rs
@@ -22,15 +22,20 @@ pub async fn remove(
 
     let args = interaction.data.options.clone();
 
-    let remove_index = args
+    let signed_remove_index = args
         .first()
         .unwrap()
         .value
         .as_ref()
         .unwrap()
-        .as_u64()
-        .unwrap() as usize;
+        .as_i64()
+        .unwrap();
 
+    if signed_remove_index.is_negative() || signed_remove_index <= 0 {
+        return create_response(&ctx.http, interaction, "Please provide an index >= 1!").await;
+    }
+
+    let remove_index = signed_remove_index as usize;
     let handler = call.lock().await;
     let queue = handler.queue().current_queue();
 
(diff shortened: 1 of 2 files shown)


src/commands/remove.rs
```diff
@@ -22,29 +22,27 @@ pub async fn remove(
...
         .value
         .as_ref()
         .unwrap()
-        .as_u64()
-        .unwrap() as usize;
+        .as_i64()
+        .unwrap();
 
+    if signed_remove_index.is_negative() || signed_remove_index <= 0 {
+        return create_response(&ctx.http, interaction, "Please provide an index >= 1!").await;
```
rafaeldamasceno, line 35: This is just a nitpick so feel free to disregard, but I think having the inequality in writing (`greater than or equal to`) would be preferable.

rafaeldamasceno (commented):
I left some comments about the wording of the bot messages.

src/events/serenity_handler.rs
```diff
@@ -81,7 +81,7 @@ impl EventHandler for SerenityHandler {
                         .create_option(|option| {
                             option
                                 .name("index")
-                                .description("Position of the track (0 is currently playing)")
+                                .description("Position of the track (next track is 1)")
```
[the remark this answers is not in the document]
afonsojramos: Agreed here!

src/commands/remove.rs
afonsojramos: I think I disagree, but it is not a hard disagree 🤔 Everyone understands math, it's not like the users do not understand a simple `>= 1` in my opinion.

src/commands/remove.rs
```diff
@@ -22,29 +22,27 @@ pub async fn remove(
...
         .unwrap()
         .value
         .as_ref()
         .unwrap()
-        .as_u64()
-        .unwrap() as usize;
+        .as_i64()
+        .unwrap();
 
+    if signed_remove_index.is_negative() || signed_remove_index <= 0 {
```
joao-conde, line 34: Just:

```rust
if !signed_remove_index.is_positive()
```

joao-conde (changes requested):

joao-conde (approved):

Issue #94, opened 2022-01-21
Queue commands stop working while a livestream is in the queue


danrpinho:
## Description

Adding a livestream (for example, a music radio) to the queue makes some commands stop working. The `/play` command stops showing which video was added to the queue (even if the video has a fixed length), while `/queue` and `/np` show error messages. The `/skip` command functions as expected, and the commands start working again after livestream playback ends.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The bot showcases the current playing song and other videos in the queue. |
| **Observed** | The bot sends an error message to the user's Discord client with the message "The application did not respond." |

## Repro Steps

1. `/play` two regular videos
2. Verify that the bot states which videos were added to the queue
3. Check that `/np` and `/queue` work properly
4. `/play` an ongoing livestream (for example, [this one](https://www.youtube.com/watch?v=5qap5aO4i9A))
5. Check that the bot only says "Searching..."
6. Verify that `/queue` throws an error message while `/np` functions as expected
7. `/play` a regular video
8. Verify that the bot does not say which video was added
9. `/skip` videos until the livestream starts
10. Verify that both `/np` and `/queue` throw error messages
11. `/skip` the livestream
12. Verify that `/play`, `/np` and `/queue` function normally (the first one with regular videos)

## Environment

| Key | Value |
| --- | --- |
| **Operating System** | Ubuntu 21.10 impish |
| **Kernel** | x86_64 Linux 5.13.0-23-generic |
| **CPU** | Intel Core i5-6500 @ 4x 3,6GHz |

danrpinho:
After poking around with the bot's behaviour with @rafaeldamasceno, we found out that it is possible that the bot doesn't know how to handle videos without a fixed duration. All commands that stop working showcase the livestream's information:

- `/np` works with a regular song while the livestream is in the queue because it only shows that video;
- Both `/np` and `/queue` stop working while the livestream is playing because they both show information about the livestream;
- `/play` commands issued while the livestream is queued/playing do not function due to the "Estimated time until play" information. This is in conformity with our observation that all these three commands function normally after livestream playback ends.

aquelemiguel:
@danrpinho Thank you for the report. 🙂
I repro'd this as well and converged on a few dangerous unwraps for `metadata.duration`. For example:

https://github.com/aquelemiguel/parrot/blob/fd2a0c08984b790f9468fb74eb37c5930abc4b59/src/commands/queue.rs#L131-L135

[...]

PR #96, opened 2022-01-21
feat: live update `/queue`

### LINKED ISSUE #92: Live update queue command
Changes to the queue (either new tracks being added or a track that finished playing) should automatically update the queue response interaction instead of the user having to click the arrow buttons to refresh it.

Visually, the queue button components could use a little change:
* Disable buttons when it's not possible to navigate to a specific page
* Use alt-codes instead of emojis
* Use primary blurple color instead of grey

## Current
image

## Expected
image

aquelemiguel:
Aims to close #92 and #94.


src/client.rs
```diff
@@ -1,11 +1,22 @@
...
     client: serenity::Client,
 }
 
+pub struct GuildQueueInteractions;
+type QueueMessage = (Message, Arc<RwLock<usize>>);
+
+impl TypeMapKey for GuildQueueInteractions {
+    type Value = HashMap<GuildId, Vec<QueueMessage>>;
+}
+
```
joao-conde, line 19: Move to its own module. In fact, wait for the approval and merge of https://github.com/aquelemiguel/parrot/pull/98 and let's merge this with the guild settings we have (possibly even rename to `GuildCache`)

src/commands/clear.rs
```diff
@@ -31,5 +32,8 @@ pub async fn clear(
         v.drain(1..);
     });
 
+    drop(handler);
+    update_queue_messages(&ctx.http, &ctx.data, &call, guild_id).await;
+
     create_response(&ctx.http, interaction, "Cleared!").await
```
joao-conde, line 38: Just an idea, maybe send the "Clear!" message first and then update the queue

src/commands/play.rs
```diff
@@ -64,6 +65,8 @@ pub async fn _play(
         EnqueueType::PLAYLIST => enqueue_playlist(&call, url).await,
     };
 
+    update_queue_messages(&ctx.http, &ctx.data, &call, guild_id).await;
```
joao-conde, line 68: Again, I would update the queue at the end of this command, I think feedback to the command is important first.

src/commands/play.rs
```diff
@@ -116,6 +119,12 @@ pub async fn _play(
 
 async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Option<Duration> {
     if !queue.is_empty() {
+        let non_livestreams = queue.iter().filter_map(|track| track.metadata().duration);
+
+        if non_livestreams.count() < queue.len() {
+            return Some(Duration::MAX);
+        }
```
joao-conde, line 126: This is not true for `/playtop` right?

src/commands/queue.rs
```diff
@@ -1,25 +1,34 @@
...
         },
     },
-    prelude::SerenityError,
+    prelude::{RwLock, SerenityError, TypeMap},
 };
-use songbird::tracks::TrackHandle;
+use songbird::{tracks::TrackHandle, Event, TrackEvent};
 
 const EMBED_PAGE_SIZE: usize = 6;
+const COLLECTION_TIMEOUT: u64 = 3600;
```
joao-conde, line 31: When this reaches 0 and it timeouts, does the embed disappear, like do we remove the embed message? If so, I would prefer `EMBED_TIMEOUT` and maybe with a little docstring specifying the time units (I know its seconds but still)

src/commands/queue.rs
```diff
@@ -48,45 +55,43 @@ pub async fn queue(
...
-                        })
+                        .add_embed(create_queue_embed(&tracks, 0))
+                        .components(|components| build_nav_btns(components, 0, num_pages))
                 })
         })
         .await?;
 
-    let message = interaction.get_interaction_response(&ctx.http).await?;
+    let mut message = interaction.get_interaction_response(&ctx.http).await?;
+    let page_lock: Arc<RwLock<usize>> = Arc::new(RwLock::new(0));
```
joao-conde, line 68: In reality, this is the `page`. Only when you call `page.read()` or `page.write()` do you get a lock. In the former a `page_rlock` in the latter a `page_wlock`

src/commands/queue.rs
```diff
@@ -97,33 +102,43 @@ pub async fn queue(
         drop(handler);
 
         let num_pages = calculate_num_pages(&tracks);
+        let mut page = page_lock.write().await;
```
joao-conde, line 105: And this would be

```rust
let mut page_wlock = page.write().await;
```

src/commands/queue.rs
```diff
@@ -141,13 +156,36 @@ fn create_queue_embed<'a>(
...
+fn build_single_nav_btn(label: &str, is_disabled: bool) -> CreateButton {
+    let mut button = CreateButton::default();
+
+    button
+        .custom_id(label.to_string().to_ascii_lowercase())
+        .label(label)
+        .style(ButtonStyle::Primary)
+        .disabled(is_disabled);
+
+    button
```
joao-conde, line 173: Is this not possible?

```suggestion
    CreateButton::default()
        .custom_id(label.to_string().to_ascii_lowercase())
        .label(label)
        .style(ButtonStyle::Primary)
        .disabled(is_disabled)
```

src/commands/queue.rs
```diff
@@ -141,13 +156,36 @@ fn create_queue_embed<'a>(
...
+}
+
+pub fn build_nav_btns(
+    components: &mut CreateComponents,
+    page: usize,
+    num_pages: usize,
+) -> &mut CreateComponents {
+    components.create_action_row(|action_row| {
+        let (cant_left, cant_right) = (page < 1, page >= num_pages - 1);
+
```
joao-conde, line 183: Nice. Except for the triggering NEW LINE *triggered*

src/commands/remove.rs
```diff
@@ -49,6 +50,10 @@ pub async fn remove(
         handler.queue().modify_queue(|v| {
             v.remove(remove_index);
         });
+
+        drop(handler);
+        update_queue_messages(&ctx.http, &ctx.data, &call, guild_id).await;
```
joao-conde, line 55: Again, maybe first send the feedback to the user ("Removed!") and then update the queue embed

src/commands/shuffle.rs
```diff
@@ -26,6 +27,9 @@ pub async fn shuffle(
         )
     });
 
+    drop(handler);
+    update_queue_messages(&ctx.http, &ctx.data, &call, guild_id).await;
```
joao-conde, line 31: Same idea, feedback to the user first.

src/commands/stop.rs
```diff
@@ -28,5 +29,8 @@ pub async fn stop(
     }
 
     queue.stop();
-    return create_response(&ctx.http, interaction, "Stopped!").await;
+    drop(handler);
+
+    update_queue_messages(&ctx.http, &ctx.data, &call, guild_id).await;
+    create_response(&ctx.http, interaction, "Stopped!").await
```
joao-conde, line 35: Feedback first

src/events/modify_queue_handler.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub http: Arc<Http>,
+    pub ctx_data: Arc<RwLock<TypeMap>>,
+    pub call: Arc<Mutex<Call>>,
+    pub guild_id: GuildId,
+}
+
+#[async_trait]
+impl EventHandler for ModifyQueueHandler {
+    async fn act(&self, ctx: &EventContext<'_>) -> Option<Event> {
+        if let EventContext::Track(_) = ctx {
```
joao-conde, line 28: In reality, you only use this handler for the `Event::Track(TrackEvent::End)` so you know the `EventContext` will be of type Track. Since you don't use anything, I think you can remove the `if let`

src/events/modify_queue_handler.rs
```diff
@@ -0,0 +1,75 @@
...
+    for (message, page_lock) in messages.iter_mut() {
+        let handler = call.lock().await;
+        let tracks = handler.queue().current_queue();
+        drop(handler);
+
+        // has the page size shrunk?
+        let num_pages = calculate_num_pages(&tracks);
+
+        let mut page = page_lock.write().await;
+        *page = usize::min(*page, num_pages - 1);
```
joao-conde, line 60: This is all to compute size shrunk, just put it together, why the new line

src/tests/utils.rs
```diff
@@ -1,29 +1,19 @@
 use serenity::model::prelude::User;
```
joao-conde, line 1: Weird that clippy does not complain, but I think you do not need this import

src/tests/utils.rs
```diff
@@ -1,29 +1,19 @@
...
-    user.name = "hello world".to_string();
-    user.discriminator = 1234;
-
-    let result = get_full_username(&user);
-    assert_eq!(result, "hello world#1234");
-}
+use crate::utils::get_human_readable_timestamp;
 
 #[test]
 fn test_get_human_readable_timestamp() {
```
joao-conde, line 7: Add a test case where it receives a `None`

src/utils.rs
```diff
@@ -28,19 +25,23 @@ pub async fn create_response(
...
+
+        if hours < 1 {
+            format!("{:02}:{:02}", minutes, seconds)
+        } else {
+            format!("{}:{:02}:{:02}", hours, minutes, seconds)
+        }
     } else {
-        format!("{}:{:02}:{:02}", hours, minutes, seconds)
+        "∞".to_string()
     }
```
joao-conde, line 45: This can better be rewritten as a match now that we have three cases: Some duration and infinite, some duration and finite, and the rest

joao-conde (changes requested):

src/commands/play.rs
aquelemiguel: Yeah, you're right. Oversight by my part, it'll show "Estimated time until play: ∞" which is incorrect.

src/commands/queue.rs
aquelemiguel: Tried it as well, it returns a `&mut CreateButton`. Could slap it a `.to_owned` but I dunno.

src/commands/queue.rs
aquelemiguel: Looks a lot worse on GitHub than VSCode, lol.

src/commands/queue.rs
aquelemiguel: The embed gets edited with `QUEUE_EXPIRED`, seen here:

https://github.com/aquelemiguel/parrot/blob/cd54d3432bf957022736a58007cf4bf20a9ea0d3/src/commands/queue.rs#L125-L135

joao-conde: Then I would say just to use `EMBED_TIMEOUT`

src/commands/queue.rs
joao-conde: I like that

joao-conde:
Haven't forgotten but will wait for merge conflicts fix 👍

src/commands/shuffle.rs
```diff
@@ -26,7 +27,11 @@ pub async fn shuffle(
         )
     });
 
-    create_response(&ctx.http, interaction, "Shuffled successfully!").await
+    drop(handler);
+
+    create_response(&ctx.http, interaction, "Shuffled successfully!").await?;
```
afonsojramos, line 32: Can you move the text to `strings.rs`?

src/commands/stop.rs
```diff
@@ -28,5 +29,9 @@ pub async fn stop(
     }
 
     queue.stop();
-    return create_response(&ctx.http, interaction, "Stopped!").await;
+    drop(handler);
+
+    create_response(&ctx.http, interaction, "Stopped!").await?;
```
afonsojramos, line 34: Can you move the text to `strings.rs`?

src/commands/clear.rs
```diff
@@ -31,5 +32,9 @@ pub async fn clear(
         v.drain(1..);
     });
 
-    create_response(&ctx.http, interaction, "Cleared!").await
+    drop(handler);
+
+    create_response(&ctx.http, interaction, "Cleared!").await?;
```
afonsojramos, line 37: Can you move the text to `strings.rs`?

afonsojramos (changes requested):

src/utils.rs
```diff
@@ -39,6 +37,51 @@ pub async fn create_embed_response(
...
+
+    let footer_text = format!(
+        "Track duration: {}\nEstimated time until play: {}",
+        get_human_readable_timestamp(metadata.duration),
+        get_human_readable_timestamp(Some(estimated_time))
+    );
+
+    embed.footer(|footer| footer.text(footer_text));
+    embed
+}
```
afonsojramos, line 83: This function should be deleted as it is now in `play.rs`.

afonsojramos (changes requested):

joao-conde (approved):

afonsojramos (approved):

PR #98, opened 2022-01-21
#42: Command `/autopause`

### LINKED ISSUE #42: Command `!autopause`
Implement a new command: `!autopause`

The listening party mode works as a toggle:  starts as `OFF` -> `!autopause` -> becomes `ON` and vice-versa

The idea is that, if the listening party mode is `ON`,  after each track, the queue stops. This allows people to discuss the track they heard without having to pause the next song, and seeking it to 00:00 when they are ready. For example, if track 1 is playing, and tracks 2 and 3 are enqueued, after track 1 finishes, nothing more plays. Only after typing `!resume` does it proceed to the next track.

joao-conde:
| - | - |
| --- | --- |
| Issue | Closes https://github.com/aquelemiguel/parrot/issues/42 |
| Dependencies | -- |
| Decisions | Usage `/autopause` as a toggle, similar to `/repeat`.<br>Make use of serenity's context cache to store `GuildSettings` such as the `autopause` flag. @aquelemiguel will be using it for other Guild-related cache.<br>Tested the edge cases of skipping and seeking to the end of the track with autopause on. |


afonsojramos (approved):
Super clean, love it 👌

aquelemiguel (approved):
Looks great, I like your refactoring of the /events dir.

A behaviour went against my intuition: skipping an autopaused song also pauses the next one. From what I've investigated, there's no simple way to check whether the track was paused before being skipped. Let's revisit this later if after using it for a while we find it that frustrating.

PR #104, opened 2022-01-23
feat: drop youtube_dl crate

### LINKED ISSUE #102: Drop `youtube_dl` crate in favour of `yt-dlp` command exec
We're only using the [`youtube_dl`](https://crates.io/crates/youtube_dl) crate for fetching the songs within a playlist.
This is overkill since we're using yt-dlp as well. Investigate executing a yt-dlp command (possibly with the `--flat-playlist` flag).

aquelemiguel:
Closes #102.

* Drops `youtube_dl` crate that was being used when enqueueing a playlist.
* Removes other crates that were being used in the lyrics command.


afonsojramos:
Maybe extracting the essential commit would be best :eyes:

PR #105, opened 2022-01-23
fix: use now playing string from `strings.rs`

afonsojramos:


src/main.rs
```diff
@@ -1,4 +1,4 @@
-use parrot::client::Client;
+use parrot::{client::Client, strings::FAIL_MAIN};
```
joao-conde, line 1: This string as a constant doesn't make much sense, because:

a) The binary is not part of the library, our `main.rs` is just the binary that uses the `parrot` library. Imagine someone using our parrot library to create their own binary, would not want this string `FAIL_MAIN` as part of the library

b) This is a server side debug string only, `strings.rs` should/is being used for message facing strings

joao-conde (changes requested):
Don't agree with the `FAIL_MAIN` string, check my other comment. The `QUEUE_NOW_PLAYING` is a good change 👍

aquelemiguel (approved):

joao-conde (approved):

PR #106, opened 2022-01-24
feat: drop youtube_dl crate

### LINKED ISSUE #102: Drop `youtube_dl` crate in favour of `yt-dlp` command exec

aquelemiguel:
Closes #102.

* Drops `youtube_dl` crate that was being used when enqueueing a playlist.
* Removes other crates that were being used in the lyrics command.


afonsojramos (commented):
Looks good code wise! Goodbye useless dependencies 👋 

Will need to test with some playlists before approving.

src/commands/play.rs
```diff
@@ -145,16 +150,19 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
+        let reader = BufReader::new(stdout);
+        let lines: Vec<String> = reader.lines().flatten().collect();
 
-        for entry in entries.iter() {
-            let url = format!(
-                "https://www.youtube.com/watch?v={}",
-                entry.url.as_ref().unwrap()
-            );
+        for line in lines.iter() {
+            let entry: Value = serde_json::from_str(line).unwrap();
```
joao-conde, line 164: Can't we get away with implicit typing here (removing the `: Value`) ?

Other than that looks good but I haven't been able to test it

aquelemiguel: Doesn't look like it, type annotations are needed in the line below.

afonsojramos (approved):
Merged with `main` so that `Cargo.lock` was updated.

Issue #97, opened 2022-01-21
Use Serenity's built-in input validation


afonsojramos:
## Rationale

Discord has input validation built into the app, however, Serenity has not released/implemented all of them, like in [create_application_command.rs#L163](https://github.com/serenity-rs/serenity/blob/cfd518e328697525339175ef6ed4a1f69ec9ba86/src/builder/create_application_command.rs#L163) which would help in #93/#95.

## Description

When [Serenity](https://github.com/serenity-rs/serenity) releases the new version, implement input validation with the built-in tools, instead of manual validation, where possible. Example: `remove` doesn't need to check if the input is `<=0`.

afonsojramos:
I think this was one of those cases that as soon as you complain, things happen, because Serenity just released [v0.10.10](https://github.com/serenity-rs/serenity/releases/tag/v0.10.10)!

I'll investigate in how many commands we can implement these new features!

PR #109, opened 2022-01-24
feat: use built-in input parsing & add `cargo.lock`

### LINKED ISSUE #103: Bump Serenity to v0.10.10
Always good to bump Serenity. Also unlocks #97.

afonsojramos:
Closes #103, #97 and #110.

image


Cargo.toml
```diff
@@ -25,7 +25,7 @@ branch = "current"
 features = ["builtin-queue", "yt-dlp"]
 
 [dependencies.serenity]
-version = "0.10"
+version = "0.10.10"
```
joao-conde, line 28: Weird, `0.10` should fetch the latest patch version, right? Which would make it `0.10.10`? Have you tried with just `0.10` but making `cargo clean` and then `cargo run` ?

aquelemiguel: Can confirm, just tested and all is working when when keeping `0.10`.

afonsojramos: Missed the `cargo clean` rolling it back!

afonsojramos: For some reason, even after a `cargo clean` it uses `v0.10.9`. Did you try `cargo clean` in this branch @aquelemiguel?

joao-conde (approved):

PR #115, opened 2022-01-25
feat: custom DJ Role for parrot commands

joao-conde:
| - | - |
| --- | --- |
| Issue | Closes https://github.com/aquelemiguel/parrot/issues/47 |
| Dependencies | `"Manage Roles"` permission is now required and Wiki should be updated after this merge. |
| Decisions | - Refactor code to make handlers smaller and call what they need.<br>- Set permissions for all guilds for all commands of the DJ Role.<br>- The DJ Role name can then be changed by the discord admins if they want, but I picked a default ("Parrot DJ").<br>- Parrot ensures the role exists by either using an already existing one or creating a new one.<br>- now playing, queue, play, summon and version are unrestrained |
| Animated GIF | dj-role-commands |

### DIFF
--- a/src/handlers/serenity.rs
+++ b/src/handlers/serenity.rs
@@ -105,6 +105,7 @@ impl SerenityHandler {
                     command
                         .name("np")
                         .description("Displays information about the current track")
+                        .default_permission(true)
                 })
                 .create_application_command(|command| {
                     command
@@ -139,7 +140,7 @@ impl SerenityHandler {
                         })
                 })
                 .create_application_command(|command| {
-                    command.name("queue").description("Shows the queue")
+                    command.name("queue").description("Shows the queue").default_permission(true)
                 })
                 .create_application_command(|command| {
                     command
@@ -204,6 +205,7 @@ impl SerenityHandler {
                     command
                         .name("version")
                         .description("Displays the current version")
+                        .default_permission(true)
                 })
         })
         .await
--- a/src/handlers/serenity.rs
+++ b/src/handlers/serenity.rs
@@ -22,8 +22,6 @@ use crate::commands::{
     remove::*, repeat::*, resume::*, seek::*, shuffle::*, skip::*, stop::*, summon::*, version::*,
 };
 
-const ROLE: &str = "Parrot DJ";
-
 pub struct SerenityHandler;
 
 #[async_trait]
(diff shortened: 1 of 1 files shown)


src/handlers/serenity.rs
```diff
@@ -17,45 +22,102 @@ use crate::commands::{
     remove::*, repeat::*, resume::*, seek::*, shuffle::*, skip::*, stop::*, summon::*, version::*,
 };
 
+const ROLE: &str = "Parrot DJ";
```
afonsojramos, line 25: nitpick: Is it possible to get the bot's name here instead of simply "Parrot DJ"?

afonsojramos (changes requested):

src/handlers/serenity.rs
```diff
@@ -17,45 +22,102 @@ use crate::commands::{
...
                     command
                         .name("pause")
                         .description("Pauses the current track")
+                        .default_permission(false)
                 })
                 .create_application_command(|command| {
                     command
                         .name("play")
                         .description("Adds a track to the queue")
+                        .default_permission(false)
```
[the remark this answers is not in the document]
afonsojramos: Summoning @aquelemiguel @rafaeldamasceno @danrpinho for opinions.

src/handlers/serenity.rs
joao-conde: The problem is that the bot's name is already its own role and can't be assigned to users. I also don't like this default but have no better idea. Thankfully it is also changeable by the Server Admin.

image

afonsojramos: I meant something like "${botName}'s DJ".

src/handlers/serenity.rs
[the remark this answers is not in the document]
afonsojramos: I mean, the development is done now, so it's a matter of simply deciding what we want, and it is a small change 🤷

src/handlers/serenity.rs
joao-conde: Seems like a good idea I will look into it

src/handlers/serenity.rs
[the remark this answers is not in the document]
joao-conde: Resolving for now. Decide in the issue and I will check the conclusions.

afonsojramos (approved):

afonsojramos:
image

Love the fact that this gets greyed out!

src/handlers/serenity.rs
```diff
@@ -116,63 +181,78 @@ impl EventHandler for SerenityHandler {
...
     }
 
-    async fn voice_state_update(
-        &self,
-        ctx: Context,
-        guild: Option<GuildId>,
-        _old: Option<VoiceState>,
-        new: VoiceState,
-    ) {
+    async fn run_command(&self, ctx: &Context, command: &mut ApplicationCommandInteraction) {
```
aquelemiguel, line 231: This isn't really a `run_command` but a more like a `register_command_callback` I'd say.

aquelemiguel (approved):
Small suggestion, but overall a high quality PR. Good stuff.

src/handlers/serenity.rs
joao-conde: No, this actually runs the command.

https://github.com/aquelemiguel/parrot/blob/jc/47-dj-permission/src/handlers/serenity.rs#L233

This actually runs the `autopause` command

afonsojramos (approved):

PR #122, opened 2022-01-27
fix: better error messaging

aquelemiguel:
image


afonsojramos (changes requested):
hmm 🤔

afonsojramos (changes requested):
I am now getting the new message instead of a `⚠️ Could not find you in any voice channel!` when I am not even connected.

image

afonsojramos (approved):
Looks good to me, and I'm not biased at all :eyes:.

PR #127, opened 2022-02-02
feat: add `forceskip` and voting system to `skip`

### LINKED ISSUE #44: Add voting system to `!skip` and introduce `!forceskip`
In larger servers, there needs to be a voting system in place for skipping songs so the queue cannot be griefed.
This means performing the following changes:

- `!skip` / `!s` should no longer immediately skip a track but instead display a new reaction-based voting system where `x` votes (maybe other metric?) skip it.
- `!forceskip` / `!fs` should be restricted to members with a **DJ permission** and immediately skips the track. This is identical to the current behavior of `!skip`.

More on this **DJ permission** on a later issue because it should also be used by other destructive commands like `!clear` and `!remove` or even `!seek`.

afonsojramos:
Closes #44.

These are the 3 situations that can happen with the `/skip` command, which I exemplified below: 

1. There are 3 users listening to the bot. One requests a skip, but it does not constitute a majority.
2. One of the users disconnected, therefore there were only two users listening to the bot, and a single vote constitutes a "majority vote" *insert PS joke here*.
3. Nothing is playing.

image


src/guild/cache.rs
```diff
@@ -10,6 +13,7 @@ type QueueMessage = (Message, Arc<RwLock<usize>>);
 #[derive(Default)]
 pub struct GuildCache {
     pub queue_messages: Vec<QueueMessage>,
+    pub current_skip_votes: Vec<UserId>,
```
aquelemiguel, line 16: This is better off being a [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html), it'll save you a few checks on the skip command.

src/commands/skip.rs
```diff
@@ -20,9 +23,43 @@ pub async fn skip(
...
     }
 
+    let mut data = ctx.data.write().await;
+    let cache_map = data.get_mut::<GuildCacheMap>().unwrap();
+
+    let cache = cache_map.entry(guild_id).or_default();
+
+    if !cache.current_skip_votes.contains(&interaction.user.id) {
+        cache.current_skip_votes.push(interaction.user.id);
+    }
```
aquelemiguel, line 35: No longer needs this check if `cache.current_skip_votes` is a HashSet instead of a vector.

src/commands/skip.rs
```diff
@@ -20,9 +23,43 @@ pub async fn skip(
...
+
+    let guild = ctx.cache.guild(guild_id).await.unwrap();
+    let skip_threshold = guild.voice_states.len() / 2;
+
+    let message = if cache.current_skip_votes.len() >= skip_threshold && queue.skip().is_ok() {
```
aquelemiguel, line 40: If I'm reading this correctly, if there are 7 users in the channel, the threshold is the integer div of 7/2=3.
So it's skipping one vote early on odd-sized lobbies, no?

aquelemiguel, line 40: I'd rather have `queue.skip()` checked for OK inside the condition, it's not very readable in the current state.
Additionally, if skipping fails, it displays the skip vote count which I'm not certain it is what you want.

src/commands/skip.rs
```diff
@@ -20,9 +23,43 @@ pub async fn skip(
...
+            SKIP_VOTE_USER,
+            skip_threshold - cache.current_skip_votes.len(),
+            SKIP_VOTE_MISSING
+        )
+    };
+
+    create_response(&ctx.http, interaction, &message).await?;
+    drop(data);
+
     Ok(())
```
aquelemiguel, line 56: No need, this gets dropped immediately. If anything do it before the create response but I wouldn't bother.

```suggestion
    create_response(&ctx.http, interaction, &message).await?
```

src/commands/skip.rs
```diff
@@ -20,9 +23,43 @@ pub async fn skip(
...
 }
+
+pub async fn forget_skip_votes(data: &Arc<RwLock<TypeMap>>, guild_id: GuildId) {
+    let mut data = data.write().await;
+    let cache_map = data.get_mut::<GuildCacheMap>().unwrap();
+
+    let cache = cache_map.get_mut(&guild_id).unwrap();
+    cache.current_skip_votes = Vec::new();
+}
```
aquelemiguel, line 65: This should probably return a Result/Option so we can cut some unwraps.

aquelemiguel (changes requested):
Quality PR, I like the choices that were made here, even if I originally envisioned interaction-based voting. This approach, while less flashy, ends up being more functional and less error-prone.

afonsojramos:
> Quality PR, I like the choices that were made here, even if I originally envisioned interaction-based voting. This approach, while less flashy, ends up being more functional and less error-prone.

Thanks! 🚀  In addition to being more functional and less error-prone, something that I didn't like about the interaction based approach is that it would make it too easy to skip something, and, in my opinion, it would be bad UX for this functionality.

I'll look at the other comments now!

src/commands/skip.rs
afonsojramos: No, because the bot is also included in the `voice_states.len()`. This was actually a doubt that I had while developing but in the end this was the correct solution. I'll exemplify:
1. 7 listeners: 8/2 > 4 votes
2. 6 listeners: 7/2 > 3 votes
3. 5 listeners: 6/2 > 3 votes

afonsojramos: In that case I won't be able to remove the last Ok(()), since I won't always have a `message`.

src/commands/skip.rs
afonsojramos: Why would we want to return an option here exactly?

src/commands/skip.rs
aquelemiguel: Ah, the bot counts! Good catch, I stand corrected.

src/commands/skip.rs
aquelemiguel: Because fetching from the context might fail so instead of unwrapping we propagate the error upwards and leave it to the caller to handle errors (or unwrap as well).

src/commands/skip.rs
aquelemiguel: That's alright!

joao-conde (approved):

aquelemiguel (approved):

Issue #50, opened 2021-10-26
Spotify support


rafaeldamasceno:
## Rationale

Most people use Spotify to listen to tracks. They also usually have playlists on Spotify. It would be nice to have support for on-the-fly conversion of Spotify links.

## Description

When a Spotify link is given, fetch the artist and title of the track (or tracks, in the case of a playlist), and then feed them into `ytdl_search` as usual.

afonsojramos:
I'm going to tackle this next, if no one opposes!

aquelemiguel:
I'm currently doing some refactoring in [`source-getters`](https://github.com/aquelemiguel/parrot/tree/source-getters) that prepares for implementations in both #15 and #50. I'd suggest either waiting or continuing work on that branch before tackling this.

Issue #15, opened 2021-10-08
SponsorBlock integration


aquelemiguel:
[SponsorBlock](https://sponsor.ajay.app/) is an open-source crowdsourced browser extension and open API for skipping sponsor segments in YouTube videos. It also keeps track of non-music segments for music videos.

The idea would be enqueueing a song, using the API to search for non-music segments (e.g., cinematic parts in music videos, credits or intermissions) and automatically skip them when the playtime enters that segment.

### Possible usage
* `!sponsorblock` as an on/off toggle for this feature

### Implementation
Triggering the seek action at specific timestamps might raise some issues.
* Should the bot periodically (every second or less) fire an event checking whether the song entered a non-music segment?
* We could also schedule events for those specific timestamps but we'd have to be extra careful adjusting the scheduling because of time altering commands like `!pause` and `!seek`.
* ...

aquelemiguel:
Just found out [yt-dlp supports SponsorBlock](https://github.com/yt-dlp/yt-dlp#sponsorblock-options) and you can remove certain categories. No need for FFmpeg gimmicks.

However, in order to run our own custom yt-dlp commands, we'd need to ditch `Restartable::ytdl` - instead download the song and pass it to `Restartable::ffmpeg`.

![image](https://user-images.githubusercontent.com/5618493/149972157-eb8a8014-2c10-4756-b8b2-672798dde926.png)

As an added bonus, supporting custom yt-dlp (or FFmpeg) commands might help us fix #88.

joao-conde:
We have to see what `Restartable::ytdl` does again, with a proper look. If it simply uses `yt-dlp` for metadata and pipes that to `Restartable::ffmpeg` then it will be easy.

aquelemiguel:
After experimenting for a few days, this revealed itself to be tougher than initially expected. This is because yt-dlp only does the Sponsorblock step after downloading the video. That's terrible because it now forces us to first download the track, trim it and then stream it (and I guess eventually delete it). Too much hassle for a small niche feature.

If we ever add playback support for local files, then we can revisit this. Closing it for now.

Issue #88, opened 2022-01-11
Unresponsiveness with long songs


danrpinho:
## Description

When playing long songs, such as radios or playlists (which usually run about an hour long), the bot suddenly stops playing audio and stops responding to commands.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The song plays until the end, and users can input commands. |
| **Observed** | The audio playback stops (around the 30-minute mark) and the bot does not respond to commands. |

## Repro Steps

1. `!play` a long video (for example, [this one](https://www.youtube.com/watch?v=syK3EZpi1sw&t=4s))
2. Listen until the video ends
3. Attempt to add commands later in the video's runtime

## Environment

| Key | Value |
| --- | --- |
| **Operating System** | Ubuntu 21.10 impish |
| **Kernel** | x86_64 Linux 5.13.0-23-generic |
| **CPU** | Intel Core i5-6500 @ 4x 3,6GHz [22.0°C] |
| **RAM** | 7866MiB / 15893MiB |

afonsojramos:
I can confirm that this happens as well. I haven't had many occasions where I listened to long videos, however, everytime that I have, it has stalled at some point.

Ps: I think this is unrelated to the Windows bug that we've discovered the other day @aquelemiguel @joao-conde

aquelemiguel:
I think it just might.
Now that I think of it, we're getting songbird directly from their current branch because of yt-dlp compatibility:

```
[dependencies.songbird]
git = "https://github.com/serenity-rs/songbird"
branch = "current"
features = ["builtin-queue", "yt-dlp"]
```

I wonder if something wrong was pushed. I'm 99.99% certain it isn't our code but a dependency.

aquelemiguel:
Thank you for the bug report, @danrpinho.

It just might, @afonsojramos, they seem too similar... 99.99% chance it's unrelated to our code however, given v0.2.0 is also failing now. Should most likely be some funky dependency on both issues. I'm still investigating this.

afonsojramos:
The thing is that your case @aquelemiguel does not fail on Linux, and this one does. If it IS the same problem, it is much worse on Windows vs Linux.

aquelemiguel:
It's the same issue, no? Only that on Windows it freezes after a minute or so and on Linux it's half an hour.

rafaeldamasceno:
[...]

PR #135, opened 2022-02-04
refactor: custom Restartables for unique sources

aquelemiguel:
Refactor towards implementations on #15 and #50.


src/sources/youtube.rs
```diff
@@ -0,0 +1,173 @@
...
+        "-f",
+        "bestaudio",
+        "-R",
+        "infinite",
+        "--no-playlist",
+        "--ignore-config",
+        "--no-warnings",
+        uri,
+        "-o",
+        "-",
```
afonsojramos, line 99: These are a lot more args than we initially had, can you go over some of them?

src/sources/ffmpeg.rs
```diff
@@ -0,0 +1,43 @@
...
+        "-f",
+        "s16le",
+        "-ac",
+        "2",
+        "-ar",
+        "48000",
+        "-acodec",
+        "pcm_f32le",
+        "-",
+    ];
```
afonsojramos, line 18: Can you go over these args? Why did we add them? Or were they already in use, but not specified by us?

src/commands/play.rs
```diff
@@ -143,20 +138,9 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
-        let reader = BufReader::new(stdout);
-        let lines: Vec<String> = reader.lines().flatten().collect();
-
-        for line in lines.iter() {
-            let entry: Value = serde_json::from_str(line).unwrap();
-            let url = entry.get("url").unwrap().as_str().unwrap().to_string();
-            enqueue_song(call, url, true, &PlayFlag::DEFAULT).await;
+    if let Some(urls) = YouTubeRestartable::ytdl_playlist(uri).await {
+        for url in urls.iter() {
+            enqueue_song(call, url.to_string(), true, &PlayFlag::DEFAULT).await;
```
afonsojramos, line 143: Much cleaner, I like it 👌

src/sources/youtube.rs
```diff
@@ -0,0 +1,173 @@
...
+            ffmpeg_yt(yt, metadata, &["-ss", &ts]).await
+        } else {
+            ffmpeg_yt(yt, metadata, &[]).await
+        }
+    }
+
+    async fn lazy_init(&mut self) -> SongbirdResult<(Option<Metadata>, Codec, Container)> {
+        _ytdl_metadata(self.uri.as_ref())
+            .await
+            .map(|m| (Some(m), Codec::FloatPcm, Container::Raw))
```
afonsojramos, line 83: This codec is for metadata only?

src/sources/youtube.rs
```diff
@@ -0,0 +1,173 @@
...
+        "bestaudio",
+        "-R",
+        "infinite",
+        "--no-playlist",
+        "--ignore-config",
+        "--no-warnings",
+        uri,
+        "-o",
+        "-",
+    ];
```
afonsojramos, line 151: The args here seem very similar to the ones in `ytdl`, maybe we could create a `generic_ytdl_args`?

src/sources/youtube.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    let youtube_dl_output = TokioCommand::new("yt-dlp")
+        .args(&ytdl_args)
+        .stdin(Stdio::null())
+        .output()
+        .await?;
```
afonsojramos, line 157: Why do we use `TokioCommand` here and `Command` in line 102?

src/sources/youtube.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    let o_vec = youtube_dl_output.stderr;
+
+    let end = (&o_vec)
+        .iter()
+        .position(|el| *el == 0xA)
+        .unwrap_or_else(|| o_vec.len());
```
afonsojramos, line 164: Can you explain this? 🤔

src/sources/youtube.rs
aquelemiguel: I pretty much copied them from the Songbird's source:
https://github.com/serenity-rs/songbird/blob/73323e58ddf47dfa2bb0e334c37e939cfbd95a86/src/input/ytdl_src.rs#L43-L55

src/sources/ffmpeg.rs
aquelemiguel: They were, please check: https://github.com/serenity-rs/songbird/blob/73323e58ddf47dfa2bb0e334c37e939cfbd95a86/src/input/ytdl_src.rs#L57-L67

afonsojramos:
Just a final comment to these changes. It seems that we are taking control of some of the things that were happening on serenity/songbird, however, I ponder whether or not that is a good thing. At first, yes, because it will allow for cleaner code and more control, but over the long run:
1. do we really want to detach ourselves so much? 
2. Do we really want to insert, into our codebase, code that we do not understand? 
3. Do we really want to have full control of ffmpeg/yt-dl args, when songbird has a way bigger community maintaining said args?

aquelemiguel:
@afonsojramos You raise some great concerns, let me provide my thoughts on this.

> Do we really want to detach ourselves so much?

I feared we'd be playing catch with Songbird and end up having to periodically check their code and reflect the changes here.
Still, the source hasn't changed for a few months now and we can always fallback to the default implementation if we wish, so I'm not too concerned here.

> Do we really want to insert, into our codebase, code that we do not understand?

Not at all. The code isn't too complex per say, but should be understood before merging.

> Do we really want to have full control of ffmpeg/yt-dl args, when songbird has a way bigger community maintaining said args?

It appears support for a `ytdl_optioned` isn't in the works for complexity reasons so the main developer suggests our approach. And I don't think it'll ever be, as handling all possible flag combinations and different outputs is too overkill and it shouldn't be Songbird's job to completely mirror yt-dlp's API.

image

image

If we really want to implement #15 or support other features in the future such as fetching an entire playlist via a video link (`--yes-playlist`), support different audio formats or download livestreams from the start (`--live-from-start`) we need to pass custom arguments to yt-dlp so this change is mandatory.

Still, I'd love to hear your thoughts on this, @afonsojramos and @joao-conde.

src/sources/youtube.rs
aquelemiguel: Not really sure, the tokio variant is just an async implementation of Command but we await regardless.
Changed to `std::process::Command` and seems to run just fine.

src/sources/youtube.rs
aquelemiguel: This is a good idea, however I'd keep it as it is for now for readability purposes. Then later make it generic if needed.

afonsojramos: Sure, I don't think it is a serious offence either way. Resolving ✅

afonsojramos:
Will try to test this out later today! 🤓

src/sources/ffmpeg.rs
joao-conde, line 18: Find out what the flags do and comment this

joao-conde (commented):
All in all, I don't understand half of what this is doing. Try to find out what some of the more complex stuff does and document it (code comments work fine) so that we actually know it (now and in the future). Until we are able to figure out every flag or instruction, I don't think we should merge. Let's document more than we usually do on this particular module because I think it is needed.

As for taking this source code out of songbird, I think it's fine. If we end up having something nice that they want we might open a PR there, but currently, let's satisfy our needs.

src/sources/ffmpeg.rs
afonsojramos: Leaving this link here: https://ffmpeg.org/ffmpeg.html#Generic-options

afonsojramos (approved):

src/sources/ffmpeg.rs
```diff
@@ -0,0 +1,43 @@
...
+        "-i",
+        "-", // read from stdout
+        "-f",
+        "s16le", // use PCM signed 16-bit little-endian format
+        "-ac",
+        "2", // set two audio channels
+        "-ar",
+        "48000", // set audio sample rate of 48000Hz
+        "-acodec",
+        "pcm_f32le", // set pcm_f32le codec
```
joao-conde, line 18: This comment adds nothing.

src/sources/youtube.rs
```diff
@@ -0,0 +1,170 @@
...
+        "--no-simulate", // ensure video is downloaded regardless of printing
+        "-f",
+        "bestaudio", // select best quality audio-only
+        "-R",
+        "infinite",        // infinite number of download retries
+        "--no-playlist",   // only download the video if URL also has playlist info
+        "--ignore-config", // disable all configuration files for a yt-dlp run
+        uri,
+        "-o",
+        "-", // stream data to terminal
```
joao-conde, line 99: "To stdout" I assume, may or may not be a terminal

src/sources/youtube.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+async fn ytdl(uri: &str) -> Result<(Child, Metadata), SongbirdError> {
+    let ytdl_args = [
+        "-j",            // print JSON information for video for metadata
+        "--no-simulate", // ensure video is downloaded regardless of printing
+        "-f",
+        "bestaudio", // select best quality audio-only
+        "-R",
+        "infinite",        // infinite number of download retries
```
joao-conde, line 94: A possible problem in the future, a maximum number, even if high, should be used I would say

src/sources/youtube.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    // this rigmarole is required due to the inner synchronous reading context
+    let stderr = yt.stderr.take();
+    let (returned_stderr, value) = task::spawn_blocking(move || {
+        let mut s = stderr.unwrap();
+        let out: SongbirdResult<Value> = {
+            let mut o_vec = vec![];
+            let mut serde_read = BufReader::new(s.by_ref());
+
+            if let Ok(len) = serde_read.read_until(0xA, &mut o_vec) {
```
joao-conde, line 117: Replace 0xA by a NEWLINE constant or something so that it is readable.

src/sources/youtube.rs
```diff
@@ -0,0 +1,170 @@
...
+async fn _ytdl_metadata(uri: &str) -> SongbirdResult<Metadata> {
+    let ytdl_args = [
+        "-j", // print JSON information for video for metadata
+        "-R",
+        "infinite",        // infinite number of download retries
+        "--no-playlist",   // only download the video if URL also has playlist info
+        "--ignore-config", // disable all configuration files for a yt-dlp run
+        uri,
+        "-o",
+        "-", // stream data to terminal
```
joao-conde, line 147: idem

src/sources/ffmpeg.rs
afonsojramos: I disagree here. These are arguments for FFMPEG, if you want to know about the FFMPEG codecs, there is a FFMPEG documentation website that you can look, the comment already says everything we need to know. At best I'd say we could add `audio`, aka `// set pcm_f32le audio codec`, since -acodec automatically scopes to Audio streams.

src/sources/youtube.rs
afonsojramos: Also, why is the comment so spaced out 🥴

src/sources/youtube.rs
afonsojramos: Additionally, can we maybe test `0xD + 0xA` as well? Just to validate that the #88 isn't related to line endings.

src/sources/youtube.rs
joao-conde: Ye, true, hence "in the future". The spacing is weird, fix that too @aquelemiguel .

src/sources/youtube.rs
joao-conde: The newline character is 0xA in both platforms. Windows just has a CR before (0xD). Proof of that is that songbird only checks for 0xA and works on both platforms.

src/sources/youtube.rs
aquelemiguel: Not my fault, fmt does that. 🤷

joao-conde (approved):

Issue #51, opened 2021-10-26
Play command flags


rafaeldamasceno:
## Rationale

When playing a playlist, you sometimes want to shuffle the playlist before playing it. Sometimes you might even want to play it right after the current track, or even interrupt the current track and skip right away. These can be all achieved via flags and it was a much appreciated feature from [Groovy](https://groovy.zendesk.com/hc/en-us/articles/360031046952-Advanced-Usage-Flags).

## Description

When using the play command, adding a flag modifies the behavior of the play command.

These were the available flags in Groovy (with the exception of **-c**hoose, which seems more appropriate as a search command). They were also available as shorthand flags with the first letter only:

- [x] **-a**ll: This instructs Groovy to queue all returned songs (...) when you queue a YouTube video in a playlist.
- [x] **-s**huffle: (...) randomize the order of songs in the playlist before it's added to the queue (...)
- [x] **-n**ext: (...) the song you queued is placed directly after the playing track in the queue. (...)
- [x] **-j**ump: Instantly jumps to the track you just queued. (...)
- [x] **-r**everse: Queue the playlist in reverse order.

rafaeldamasceno:
I'd like to return to this feature, since it's the one I miss the most and I want something to start out with the code as well. Do you think it makes sense to keep the flags in the query, or should I make these optional booleans in the slash command?

afonsojramos:
Definitely adding these as optional booleans is the way to go moving forward for any additional arguments imo!

aquelemiguel:
@rafaeldamasceno I reckon this is a good chance to use [select menus](https://discord.com/developers/docs/interactions/message-components#select-menus) as they support both single and multiple selections.
Was it possible to use multiple flags on Groovy?

rafaeldamasceno:
> Was it possible to use multiple flags on Groovy?

Yes, it was.

> @rafaeldamasceno I reckon this is a good chance to use [select menus](https://discord.com/developers/docs/interactions/message-components#select-menus) as they support both single and multiple selections.

Yes and no... Do we want to break the current flow of just posting the query and having the bot reply? We'd need to have this intermediate step to select the flags (which I don't even know if is possible to do in the same slash command as the one that receives the query), or create a new command for using the flags. None of these options seem good to me.

afonsojramos:
@rafaeldamasceno have you tested it out? I'm pretty sure it is pretty seamless using discord's built-in optionals and stuff

rafaeldamasceno:
[...]

PR #140, opened 2022-02-13
feat: `/play` subcommands (next, all, reverse, shuffle)

StaticRocket:
Setup basic play flags leveraging current PlayFlag enum system. Hopefully a good starting point for #51 

This adds the flags for:
- `PLAYALL` which overrides the URL logic to force the playlist to unravel for any URL that ytdl can interpret as a playlist
- `PLAYTOP` which replaces the current /playtop command

Please let me know if anything looks odd. This is my first real attempt at rust.


src/handlers/serenity.rs
```diff
@@ -134,18 +134,14 @@ impl SerenityHandler {
...
                             option
-                                .name("query")
-                                .description("The media to play")
+                                .name("flag")
+                                .description("Play command flag")
                                 .kind(ApplicationCommandOptionType::String)
-                                .required(true)
+                                .add_string_choice("Place this song at the top of the queue", "top")
+                                .add_string_choice("Place all songs in the playlist in the queue", "all")
+                                .required(false)
```
aquelemiguel, line 144: We could make use of [subcommands](https://discord.com/developers/docs/interactions/application-commands#example-walkthrough) here for the flags. Because right now this is a little finicky to interact with:

image

image

This grouping would be a little clearer and would allow for specific help text (e.g. **/play query top** with helper text **"Place track at the top of the queue"**):

image

aquelemiguel (changes requested):
@StaticRocket Thank you for your submission. ☺️

Functionality-wise, it seems great. I just requested a few changes regarding subcommands, but they're open for discussion if anyone wants to chime in with an even better idea.

However, please be aware that #51 is already assigned to @rafaeldamasceno.

Before proceeding with this, let's understand whether these changes are compatible with the work Rafael might've already done. Let's respect the assignments. 🙂

src/handlers/serenity.rs
afonsojramos: I don't know if I agree with you here @aquelemiguel, since this would pollute the command listing a bit. Additionally, the flag is optional, so you don't even need to type it.

joao-conde: I like the suggested approach much more too.

aquelemiguel: @afonsojramos A few considerations: 
* Frequently used commands always show up on top and there's autocomplete (plus we already had **/playtop** as a separate command and it was okay).
* UX-wise currently this is really awkward to interact with. Final command text looks weird and too long (if anything, it should just be `flag: top`) and there's no place for helper text.
* This is the expected use-case for subcommands.
 
I recommend pulling the PR and messing with it for a while.

StaticRocket: Yeah, I agree with the remark about the helper text. That alone makes me want to change the implementation, but functionally I think I still prefer the flags.

Its quick and optional, but I wonder if there's some standard already in place to leverage instead of debating this single use case.

aquelemiguel: > Yeah, I agree with the remark about the helper text. That alone makes me want to change the implementation, but functionally I think I still prefer the flags.

Yup, I think we can all agree flags are faster for old prefix-based commands.

> Its quick and optional, but I wonder if there's some standard already in place to leverage instead of debating this single use case.

For slash commands I'm pretty certain subcommands are the standard, please check [the docs](https://discord.com/developers/docs/interactions/application-commands#subcommands-and-subcommand-groups).
@afonsojramos Discord might nest subcommands in the future:

image

joao-conde: And flags actually seem slower to use than a specific subcommand. Flags as a concept are useful in CLI applications, for slash commands I don't see it being coherent or useful. I much prefer the subcommand approach.

StaticRocket: Alright, I'll start working on a subcommand implementation, but as unfamiliar with rust as I am, I am even more unfamiliar with serenity's subcommand system so I'll need another day or two. Are there any other things I should look into? Everyone happy with leveraging the PlayFlag enum?

Should we continue to let the interaction attributes override the command arguments or should we make that only work if the argument isn't explicitly given? Or should the argument that are being overridden be removed? I'm sure I'll have some input here when I go to implement it, but for future functionality is there anything that comes to mind?

StaticRocket:
Well this all works now, but there's an issue with the way the playlist spooling system works right now. Queuing an instant mix will result in an infinite queue, adding results until the bot is restarted. This continues even if the bot is forced to leave, or at least the yt-dlp thread continues after the bot is forced to leave. The queue seems to stop growing after the `/leave` command is issued.

joao-conde:
@StaticRocket I am unfamiliar with what an instant mix is. If you can fill me in I can propose some solution. 

I know that for live streams we check the track metadata for the duration of the track.

In any case, I will review once there are no known bugs/issues with the PR.

aquelemiguel:
@StaticRocket I'm also unfamiliar with what instant mixes are, but I'm assuming it's like those infinitely generating playlists Spotify has? Could you please provide an example link?

StaticRocket:
Sure, here's a link to a YouTube instant mix: https://youtu.be/1V_xRb0x9aw?list=RD1V_xRb0x9aw

And what they look like:
image

The only reason we haven't had to deal with this issue before now is because these don't quite work like a playlist. They cannot be accessed by the usual playlist formatted URL.

https://www.youtube.com/playlist?list=RD1V_xRb0x9aw

But you'll find that when you reach the last song in the playlist the size will increase as it adds more videos.

This is probably something we should use those new ytdl parameters for, but I'm not sure what we should do. Having a hard cutoff for playlist lengths seems bad. Having something like a rate-limit that only applies to instant mixes seems alright, but we still have to deal with the runaway process, assuming that wasn't just my machine being weird.

I assume the ideal thing would be to wait until the current video is half way done before processing the next video in the mix (making this a new enqueuing method just for mixes and infinite length playlists), but I haven't looked to see how feasible that is yet.

joao-conde:
So a mix is kind of youtube suggestions, in a playlist, but when you reach the end it just gives you new ones? Ok, if that is so, I would say we either:

a) don't allow queueing mixes
b) queue a frozen snapshot of the mix, meaning we take the first tracks and queue them, ignoring newly added ones

I will take a look at the source later to suggest such an implementation but I believe it would be possible.

aquelemiguel:
@StaticRocket I just checked and yt-dlp seems to return 256 songs with an instant mix. I wouldn't worry much about this, TBH.

Alternatively, the simplest fix would be limiting the number of songs, either by trimming yt-dlp's output on the play command or by passing a custom yt-dlp argument (possibly **--playlist-end**). But again, I personally wouldn't bother, I don't know what the rest think.

StaticRocket:
I'm comfortable with leaving it, but currently the /stop and /clear commands can't do anything to stop the queueing until all 256 entries have been added.

Not sure if people will like that.

aquelemiguel (commented):
This is working pretty well! ✨
Before merging, don't forget to run `cargo fmt` so our pipeline passes.

aquelemiguel:
> Currently the /stop and /clear commands can't do anything to stop the queueing until all 256 entries have been added.

@StaticRocket This is true (for any playlist really) and not ideal. Because AFAIK communication between commands can only be accomplished through `context.data`, I can only really see a solution that would involve reading some sort of "stop flag". Set it to true in `/stop` and have `/play` break the queue loop if it's set.

I don't think this is super relevant to this issue in particular but definitely something to think about in the future.

src/handlers/serenity.rs
```diff
@@ -125,27 +125,46 @@ impl SerenityHandler {
...
+                        .description("Add a track to the queue")
                         .default_permission(true)
                         .create_option(|option| {
                             option
-                                .name("query")
-                                .description("The media to play")
-                                .kind(ApplicationCommandOptionType::String)
-                                .required(true)
+                                .name("default")
+                                .description("Place this song at the bottom of the queue")
```
aquelemiguel, line 133: Sorry, just a final change, I missed this. 😓

Please rename default to **end** and change description to **"Add a track to the end of the queue"** - it'll make it more in line with the rest of the descriptions.

aquelemiguel (changes requested):

aquelemiguel (approved):
@StaticRocket LGTM, thank you for your work and insights. 🥰

src/commands/play.rs
```diff
@@ -21,15 +21,17 @@ pub async fn play(
     ctx: &Context,
     interaction: &mut ApplicationCommandInteraction,
 ) -> Result<(), SerenityError> {
-    _play(ctx, interaction, &PlayFlag::DEFAULT).await
+    _play(ctx, interaction, &PlayFlag::END).await
```
joao-conde, line 24: Now that we are it, let's change the `_play` signature to take `flag: PlayFlag` instead of a reference (does not really make sense to pass a reference for an enum value)

src/commands/play.rs
```diff
@@ -40,6 +42,12 @@ pub async fn _play(
         .as_str()
         .unwrap();
 
+    let flag = match subcommand_args.name.as_str() {
+        "next" => &PlayFlag::NEXT,
+        "all" => &PlayFlag::ALL,
+        _ => flag,
+    };
```
joao-conde, line 49: make the default case return the `PlayFlag::END` maybe (and for clarity add a "end" => PlayFlag::END" too)

src/commands/play.rs
```diff
@@ -50,12 +58,17 @@ pub async fn _play(
...
+        PlayFlag::ALL => EnqueueType::PLAYLIST,
+        _ => {
+            if url.contains("youtube.com/playlist?list=") {
+                EnqueueType::PLAYLIST
+            } else if url.starts_with("http") {
+                EnqueueType::URI
+            } else {
+                EnqueueType::SEARCH
+            }
+        }
```
joao-conde, line 71: Can be rewritten as:

```rust
_  if url.contains("youtube.com/playlist?list=") => EnqueueType::PLAYLIST,
_  if url.starts_with("http") => EnqueueType::URI,
_ => EnqueueType::SEARCH
```

src/commands/play.rs
```diff
@@ -75,13 +88,13 @@ pub async fn _play(
...
             EnqueueType::URI | EnqueueType::SEARCH => match flag {
-                PlayFlag::PLAYTOP => {
+                PlayFlag::NEXT => {
                     let track = queue.get(1).unwrap();
                     let embed = create_queued_embed(PLAY_TOP, track, estimated_time).await;
 
                     edit_embed_response(&ctx.http, interaction, embed).await?;
                 }
-                PlayFlag::DEFAULT => {
+                _ => {
```
joao-conde, line 97: This is the PlayFlag::END right? Make it explicit.

src/commands/mod.rs
```diff
@@ -18,8 +17,9 @@ pub mod summon;
 pub mod version;
 
 pub enum PlayFlag {
-    DEFAULT,
-    PLAYTOP,
+    END,
+    NEXT,
+    ALL,
```
joao-conde, line 22: While we are at it, refactor this enum to be something like `PlayMode` what do you think @aquelemiguel ? Also, usually enum values are PascalCase (so All, End, Next).

src/commands/play.rs
```diff
@@ -133,14 +147,13 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
         }
-        PlayFlag::PLAYTOP => Some(top_track_duration - top_track_elapsed),
     }
 }
 
 async fn enqueue_playlist(call: &Arc<Mutex<Call>>, uri: &str) {
     if let Some(urls) = YouTubeRestartable::ytdl_playlist(uri).await {
         for url in urls.iter() {
-            enqueue_song(call, url.to_string(), true, &PlayFlag::DEFAULT).await;
+            enqueue_song(call, url.to_string(), true, &PlayFlag::END).await;
```
joao-conde, line 156: Change `enqueue_song` too to take a owned value of the play flag instead of a reference.

joao-conde (changes requested):
@aquelemiguel tell me what you think of these changes if you agree to.

@StaticRocket let me know if anything is unclear. Don't forget to run `cargo fmt` and `cargo clippy`

src/commands/mod.rs
aquelemiguel: Yeah, I like it, let's drop the flag naming.

aquelemiguel:
@joao-conde Agree with all your takes! 🚀

src/commands/play.rs
StaticRocket: Yeah, I should do that. For some reason I thought I should allow for a case where the passed flag takes priority, in case some command decides to wrap the play command. In hindsight that functionality should be removed as the play command is now very dependent on message context.

StaticRocket: @joao-conde something I forgot about this is that it currently covers the '&_' case. Modifying this to specifically include the "end" case will make the pattern non-exhaustive.

src/commands/mod.rs
afonsojramos: Circling back to this conversation due to the PascalCase enums. In Rust, Enums are in Pascal case, which we can already find in our code in the `Connection` enum. However, it is true that the old `EnqueueType` and `PlayFlag` were all capitalised... Either way, we should move to the standard imo.

aquelemiguel:
@StaticRocket You're aiming to implement the rest of the subcommands on this PR? I
If so, I'll leave it open.

src/handlers/serenity.rs
```diff
@@ -166,6 +166,32 @@ impl SerenityHandler {
                                         .required(true)
                                 })
                         })
+                        .create_option(|option| {
+                            option
+                                .name("reverse")
+                                .description("Add all tracks in reverse if the URL refers to a video and a playlist")
```
aquelemiguel, line 172: Not quite, should work for both video/playlist links and simple playlist links.
For simplicity's sake, reword this to **"Add a playlist to the queue in reverse order"**.

src/handlers/serenity.rs
```diff
@@ -166,6 +166,32 @@ impl SerenityHandler {
...
+                                        .name("query")
+                                        .description("The media to play")
+                                        .kind(ApplicationCommandOptionType::String)
+                                        .required(true)
+                                })
+                        })
+                        .create_option(|option| {
+                            option
+                                .name("shuffle")
+                                .description("Add all tracks in random order if the URL refers to a video and a playlist")
```
aquelemiguel, line 185: And here **"Add a playlist to the queue in random order"**.

src/commands/play.rs
```diff
@@ -190,19 +190,28 @@ async fn create_queued_embed(
...
+    };
+
+    // safeguard against ytdl dying on a private / deleted video and killing the playlist
+    let source = match source_return {
+        Ok(source) => source,
+        Err(error) => {
+            println!("{}", error);
+            return;
+        }
     };
```
aquelemiguel, line 207: Good catch! 🚀

aquelemiguel (changes requested):

afonsojramos (changes requested):

src/commands/play.rs
```diff
@@ -150,10 +140,20 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
-        for url in urls.iter() {
-            enqueue_song(call, url.to_string(), true, &PlayFlag::END).await;
+        let ordered_urls = match flag {
+            PlayMode::REVERSE => urls.iter().rev().cloned().collect(),
+            PlayMode::SHUFFLE => {
+                let mut urls_copy = urls.clone();
+                let mut rng = thread_rng();
+                urls_copy.shuffle(&mut rng);
+                urls_copy
+            }
```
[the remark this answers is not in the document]
afonsojramos: I like this implementation actually, it looks very clean for what it is. What are the benefits of one over the other @aquelemiguel?

src/commands/play.rs
afonsojramos: Indeed! But maybe we could add some text to explain what happened?

src/commands/play.rs
```diff
@@ -133,14 +137,23 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
-        for url in urls.iter() {
-            enqueue_song(call, url.to_string(), true, &PlayFlag::DEFAULT).await;
+        let ordered_urls = match flag {
+            PlayMode::REVERSE => urls.iter().rev().cloned().collect(),
+            PlayMode::SHUFFLE => {
+                let mut urls_copy = urls.clone();
+                let mut rng = thread_rng();
+                urls_copy.shuffle(&mut rng);
+                urls_copy
+            }
```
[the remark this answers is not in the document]
aquelemiguel: Yeah, I suggested that in my review above, it's cleaner to let yt-dlp do the playlist manipulation work.

rafaeldamasceno:
@StaticRocket Thank you for your great work on these commands!

As for the feature and as a discussion topic for everyone, I have some thoughts I want to add, even though I have probably missed this train. When I first suggested the flags feature, I had in mind the possibility of mixing and matching flags (for example, shuffling a playlist and playing it next with -sn, playing all the videos in a video with playlist instantly with -aj, etc.). 

Is there any plan for mixing and matching flags that doesn't result in a subcommand explosion?

src/commands/play.rs
aquelemiguel: @afonsojramos You mean a Discord message to inform the user?
Something like **"An error occurred while trying to queue track `XXXXX`! This video might be privated."**?

src/commands/play.rs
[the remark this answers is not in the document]
aquelemiguel: @afonsojramos This block is just reordering the playlist but we ask yt-dlp to return it already correctly ordered.
So we can delete this code and pass this transformation overhead to yt-dlp (which is better, even if negligible).

aquelemiguel:
> Is there any plan for mixing and matching flags that doesn't result in a subcommand explosion?

I thought about that too, but I believe right now this isn't possible with slash commands. The optional flags approach was finicky even with only one flag and with subcommands you'd have to cover all possible flags combinations.

I think here we should try to cover the most popular cases and think of this less like flags and more like subcommands. It's better to have a few stable, intuitive subcommands instead of a dodgy fully flexible flag implementation is what I mean. It does suck to lose some functionality, but following Discord standards benefits us in the long run.

afonsojramos:
> I think here we should try to cover the most popular cases and think of this less like flags and more like subcommands. It's better to have a few stable, intuitive subcommands instead of a dodgy fully flexible flag implementation is what I mean. It does suck to lose some functionality, but following Discord standards benefits us in the long run.

Agree with @aquelemiguel here.

rafaeldamasceno:
> It's better to have a few stable, intuitive subcommands instead of a dodgy fully flexible flag implementation is what I mean.

While I undestand what you mean, I don't believe the latter option would be dodgy or unintuitive. If manually parsing a flags string parameter is undesirable (and I can agree with that), having the flags as optional booleans in the play command would still allowed users to know what they were used for and for them to activate only the ones they wanted to use. 

The main disadvantage I see with the subcommands approach is that it doesn't lose *some* functionality, but instead most of the flexibility that the flags system would provide. Currenly, I can do all of the behavior of these commands by manually doing the work after queueing something, and since they only provide one functionality, it's not that much work (i.e. I'm not sure how much additional value these subcommands provide, since they only remove the need of calling a single command afterwards). The purpose of the flags would be to have any of the desired behaviors to be done automatically when enqueueing someting. I understand that more subcommands can be added in the future (some for shuffling and playing the playlist next, or any other combinations that are deemed common or useful), but IMHO it's only a matter of time before a combination that wasn't implemented is requested because it's used a lot by a specific group or person.

aquelemiguel:
> If manually parsing a flags string parameter is undesirable

@rafaeldamasceno Something like `/play query flags -a -r` where flags is a string option?
It mixes application commands and CLI interaction but I don't think it's atrocious. 🤷 

> having the flags as optional booleans in the play command

This would look terrible, no? For example, `/play query all true random true`. Options must always have a value.

> Currently, I can do all of the behavior of these commands by manually doing the work after queueing something, and since they only provide one functionality, it's not that much work

Sort of, play with the random flag is similar to playing and then shuffling, but if you had other tracks in the queue, they'd also be mixed with the playlist. But I see what you mean, they're not fantastic shortcuts.

> I understand that more subcommands can be added in the future (some for shuffling and playing the playlist next, or any other combinations that are deemed common or useful), but IMHO it's only a matter of time before a combination that wasn't implemented is requested because it's used a lot by a specific group or person.

I agree, but it's a little difficult for me to picture how useful these combinations are because I never really used flags on other bots. 
Let's ensure here on #140 that all the flags work well separately via subcommands. Then I suggest opening another issue for discussing how we'd go about mixing and matching flags in an intuitive way.

rafaeldamasceno:
@afonsojramos For sure, it's just that this PR isn't really aligned with what I was expecting in #51 and the discussion then started here. I will continue the discussion there.

In any case, @aquelemiguel the next steps you proposed sounds great. What I then recommend is that #51 isn't closed by this PR (although very advanced by it) and that this development takes into account ways to make it easy to invoke internal functions with different flags (e.g `enqueue_song` was changed to receive a single `PlayMode` and matches only on one `PlayMode` when there could be multiple, `PlayMode` could indeed just be called `PlayFlag` again, etc.). Otherwise the current rework of the playing/enqueing will need to be undone almost straight away when implementing multiple flags.

src/commands/play.rs
joao-conde: Ye, use both, that's what I meant, just for clarity, like

```rust
"end" => PlayMode::End,
_ => PlayMode::End
```

src/commands/play.rs
joao-conde: The println right now does not make much sense, as this prints to the server side and we don't have server logs (which in the future could be something we look into, using a logger).

Sending a message for each private track to the discord server might be too much. I would say just fail silently (return only).

afonsojramos: I actually would leave the print so we can easily replace with the logger in the future.

afonsojramos (approved):
Approving since the PR is already in a mergeable state. We can update the Shuffle and Reverse changes in another PR or in this one, up to you @StaticRocket, in my opinion.

aquelemiguel:
> Approving since the PR is already in a mergeable state. We can update the Shuffle and Reverse changes in another PR or in this one, up to you @StaticRocket, in my opinion.

Sure, that and the jump subcommand is missing but can also be tackled in a different one.

StaticRocket:
> the jump subcommand is missing

Absolutely going to leave that to someone else. It's not a bad feature, but I don't see myself using it that much. Implementation wise it should be easy; just make a call to `/stop` before queuing anything.

src/commands/play.rs
```diff
@@ -50,45 +53,45 @@ pub async fn _play(
...
 
     let call = manager.get(guild_id).unwrap();
 
     match enqueue_type {
-        EnqueueType::URI => enqueue_song(&call, url.to_string(), true, flag).await,
-        EnqueueType::SEARCH => enqueue_song(&call, url.to_string(), false, flag).await,
-        EnqueueType::PLAYLIST => enqueue_playlist(&call, url).await,
+        EnqueueType::Link => enqueue_song(&call, url.to_string(), true, &mode).await,
+        EnqueueType::Search => enqueue_song(&call, url.to_string(), false, &mode).await,
+        EnqueueType::Playlist => enqueue_playlist(&call, url, &mode).await,
```
joao-conde, line 68: `enqueue_song`/`enqueue_playlist` should take an owned value for the mode.

src/commands/play.rs
```diff
@@ -50,45 +53,45 @@ pub async fn _play(
...
     };
 
     let handler = call.lock().await;
     let queue = handler.queue().current_queue();
     drop(handler);
 
     if queue.len() > 1 {
-        let estimated_time = calculate_time_until_play(&queue, flag).await.unwrap();
+        let estimated_time = calculate_time_until_play(&queue, &mode).await.unwrap();
```
joao-conde, line 76: Make `calculate_time_until_play` take ownership of the mode. You will get "moved value" errors but you can derive the Copy trait for these enums since they are so "light".

src/commands/play.rs
```diff
@@ -50,45 +53,45 @@ pub async fn _play(
...
 
         match enqueue_type {
-            EnqueueType::URI | EnqueueType::SEARCH => match flag {
-                PlayFlag::PLAYTOP => {
+            EnqueueType::Link | EnqueueType::Search => match mode {
+                PlayMode::Next => {
```
joao-conde, line 80: Not sure, but try matching on a tuple, like:

```rust
match (enqueue_type, mode) {
  (EnqueueType::Link | EnqueueType::Search, PlayMode::Next) => ...
}
```

src/commands/play.rs
```diff
@@ -103,7 +106,7 @@ pub async fn _play(
     Ok(())
 }
 
-async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Option<Duration> {
+async fn calculate_time_until_play(queue: &[TrackHandle], mode: &PlayMode) -> Option<Duration> {
```
joao-conde, line 109: Take ownership of mode.

src/commands/play.rs
```diff
@@ -133,14 +137,13 @@ async fn calculate_time_until_play(queue: &[TrackHandle], flag: &PlayFlag) -> Op
...
         }
-        PlayFlag::PLAYTOP => Some(top_track_duration - top_track_elapsed),
     }
 }
 
-async fn enqueue_playlist(call: &Arc<Mutex<Call>>, uri: &str) {
-    if let Some(urls) = YouTubeRestartable::ytdl_playlist(uri).await {
-        for url in urls.iter() {
-            enqueue_song(call, url.to_string(), true, &PlayFlag::DEFAULT).await;
+async fn enqueue_playlist(call: &Arc<Mutex<Call>>, uri: &str, mode: &PlayMode) {
```
joao-conde, line 143: Take ownership of mode.

src/commands/play.rs
```diff
@@ -177,19 +180,25 @@ async fn create_queued_embed(
     embed
 }
 
-async fn enqueue_song(call: &Arc<Mutex<Call>>, query: String, is_url: bool, flag: &PlayFlag) {
-    let source = if is_url {
-        YouTubeRestartable::ytdl(query, true).await.unwrap()
+async fn enqueue_song(call: &Arc<Mutex<Call>>, query: String, is_url: bool, mode: &PlayMode) {
```
joao-conde, line 183: Idem

joao-conde (changes requested):
LGTM, make the small changes requested and I approve it.

aquelemiguel (approved):
Reviewed with @joao-conde, LGTM as well. ☺️

afonsojramos (approved):

afonsojramos:
Awesome PR, and awesome contributions @StaticRocket!! 🚀

joao-conde (approved):
LGTM

PR #143, opened 2022-02-16
fix: queue messages not live updating with playlist enqueue

### LINKED ISSUE #141: `/queue` isn't live updating with playlist enqueue
Related to #92.
The queue command only updates when the entire playlist has been enqueued instead refreshing with each song.

Reproduce by calling `/queue` followed by `/play PLAYLIST_LINK`. It should only refresh the queue message after all songs have been queued.

There's already an `update_queue_messages(http, data, call, guild_id)` method to help with this.
Should just be a matter of calling it between tracks.

aquelemiguel:
Closes #141.

### DIFF
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -74,20 +74,20 @@ pub async fn play(
     match mode {
         PlayMode::End => match query_type {
             QueryType::Keywords | QueryType::VideoLink => {
-                enqueue_song(&ctx, &call, guild_id, url.to_string(), query_type).await;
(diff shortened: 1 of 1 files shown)


afonsojramos:
Here I was thinking that this was one of those easy issues that no one takes in purpose to promote help from outside contributors 😂  Marking with the label `🐤 good first issue` was just for the giggles then, I see :eyes:

Will test later as the crux of this issue is how it handles the messaging.

joao-conde (approved):

afonsojramos (approved):

PR #144, opened 2022-02-16
feat: play jump subcommand

aquelemiguel:
* Implements `/play jump` subcommand as described by #51.
* Adds previously missing behaviours like front-pushing a playlist.
* Refactors the play logic to be much more readable.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -16,7 +16,7 @@ pub mod stop;
 pub mod summon;
 pub mod version;
 
-#[derive(Copy, Clone)]
+#[derive(Clone, Copy)]
 pub enum PlayMode {
     End,
     Next,
@@ -26,7 +26,7 @@ pub enum PlayMode {
     Jump,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Clone, Copy)]
 pub enum QueryType {
     Keywords,
     VideoLink,
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -23,10 +23,12 @@ pub enum PlayMode {
     All,
     Reverse,
     Shuffle,
+    Jump,
 }
 
-pub enum EnqueueType {
-    Link,
-    Search,
-    Playlist,
+#[derive(Copy, Clone)]
+pub enum QueryType {
+    Keywords,
+    VideoLink,
+    PlaylistLink,
 }
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -67,6 +67,10 @@ pub async fn play(
 
     let call = manager.get(guild_id).unwrap();
 
+    let handler = call.lock().await;
+    let queue_was_empty = handler.queue().is_empty();
+    drop(handler);
+
     match mode {
         PlayMode::End => match query_type {
             QueryType::Keywords | QueryType::VideoLink => {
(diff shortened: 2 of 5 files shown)


src/commands/mod.rs
```diff
@@ -23,10 +23,12 @@ pub enum PlayMode {
...
     Reverse,
     Shuffle,
+    Jump,
 }
 
-pub enum EnqueueType {
-    Link,
-    Search,
-    Playlist,
+#[derive(Copy, Clone)]
```
joao-conde, line 29: I didn't want to be petty in the last guy's PR, but order alphabetically.

joao-conde (changes requested):

src/commands/play.rs
```diff
@@ -53,20 +57,68 @@ pub async fn play(
...
+        },
+        PlayMode::All | PlayMode::Reverse | PlayMode::Shuffle => match query_type {
+            QueryType::VideoLink | QueryType::PlaylistLink => {
+                enqueue_playlist(&call, url, mode).await;
+            }
+            _ => {
+                edit_response(&ctx.http, interaction, PLAY_ALL_FAILED).await?;
+                return Ok(());
+            }
+        },
```
afonsojramos, line 120: I know the user isn't "supposed to", but if he inputs a single youtube video with this command, the parrot gets stuck.
image
image

So maybe I would check for `list=` in the case of a `QueryType::VideoLink`.

src/commands/play.rs
```diff
@@ -138,12 +216,18 @@ async fn calculate_time_until_play(queue: &[TrackHandle], mode: PlayMode) -> Opt
...
-async fn enqueue_playlist(call: &Arc<Mutex<Call>>, uri: &str, mode: PlayMode) {
+async fn enqueue_playlist(
+    call: &Arc<Mutex<Call>>,
+    uri: &str,
+    mode: PlayMode,
+) -> Option<Vec<String>> {
     if let Some(urls) = YouTubeRestartable::ytdl_playlist(uri, mode).await {
-        for url in urls {
-            enqueue_song(call, url.to_string(), true, mode).await;
+        for url in urls.iter() {
```
afonsojramos, line 225: Why did we add the `.iter()` here?

afonsojramos (changes requested):
Enjoyed the splitting of responsibilities in the play command, looks almost like how we organised `serenity.rs` 👌 

Just left two comments, one that truly needs to be fixed.

src/commands/play.rs
StaticRocket: It's been about 2 years since I wrote my first music bot, and it was in a completely different language but I vaguely remember a flag to make ytdl always return a list of results, even if it only contains a single element. Not saying that it's a better solution, but it won't break Soundcloud playlist support. (This may not even be an option anymore, though.)

afonsojramos: That is true, but I suppose that they don't face the same issue where we have playlists inside normal video links.

StaticRocket: Wait, I've got a trick. We're just collecting URLs with the playlist command to then feed back into yt-dlp later, right? Why not use the "webpage_url" attribute instead of just "url" ? The "webpage_url" exists for both playlist and single video queries, and it yields a value that yt-dlp can resolve later.

src/commands/play.rs
aquelemiguel: Because `urls` is now returned and the loop was taking ownership.

src/commands/play.rs
joao-conde: If yt-dlp can resolve that would be cool. Let's check that out @aquelemiguel

afonsojramos: Awesome suggestion indeed 👀

afonsojramos (approved):
Autobots, transform and rollout!

joao-conde (approved):

PR #150, opened 2022-02-17
fix: enqueue playlists

### LINKED ISSUE #147: Any `/play` subcommand causes panic when enqueuing a playlist
## Description

When using any `/play` subcommand with a playlist, a panic occurs.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The playlist is queued. |
| **Observed** | Nothing happens. |

## Repro Steps

1. Use any  `/play` subcommand with a playlist.
2. Observe nothing happened by running `/queue` after a while.

## Environment

Key | Value
-- | --
Operating System | Ubuntu 21.10 impish
Kernel | x86_64 Linux 5.13.0-28-generic
CPU | Intel Core i5-6500 @ 4x 3,6GHz
RAM | 15893MiB

## Screens

```
fev 17 14:11:21 rho parrot[219829]: thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code', src/commands/play.rs:281:36
```

afonsojramos:
Closes #147.

### DIFF
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -276,9 +276,10 @@ async fn enqueue_song(
     query_type: QueryType,
 ) {
     let source_return = match query_type {
-        QueryType::VideoLink => YouTubeRestartable::ytdl(query, true).await,
+        QueryType::VideoLink | QueryType::PlaylistLink => {
+            YouTubeRestartable::ytdl(query, true).await
+        }
         QueryType::Keywords => YouTubeRestartable::ytdl_search(query, true).await,
-        QueryType::PlaylistLink => unreachable!(),
     };
 
     // safeguard against ytdl dying on a private/deleted video and killing the playlist
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -77,7 +77,7 @@ pub async fn play(
                 enqueue_song(ctx, &call, guild_id, url.to_string(), query_type).await;
             }
             QueryType::PlaylistLink => {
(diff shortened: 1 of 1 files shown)


src/commands/play.rs
```diff
@@ -276,9 +276,10 @@ async fn enqueue_song(
     query_type: QueryType,
 ) {
     let source_return = match query_type {
-        QueryType::VideoLink => YouTubeRestartable::ytdl(query, true).await,
+        QueryType::VideoLink | QueryType::PlaylistLink => {
```
joao-conde, line 279: In the context of an `enqueue_song` how can a Playlist link get here? The `enqueue_ playlist` handles playlist links, not `enqueue_song`. My point being, this does not seem the correct way to fix it (even tho it does fix it)

aquelemiguel: This fix doesn't seem right, it should remain unreachable because it doesn't make sense for `enqueue_song` to receive playlist links. Given that the output of the `YouTubeRestartable::ytdl_playlist` always is an array of links, the fix should be passing `QueryType::VideoLink` to the `enqueue_song` method inside the loop.

Can you test this?

aquelemiguel (changes requested):

src/commands/play.rs
afonsojramos: > I think we should just hardcode pass as a parameter the query type to VideoLink right? Is there any case where ytdl won't return a single link/url ?

This was the other approach that I considered, and if you think it fits more, I'll make such a change.

aquelemiguel (approved):

Issue #149, opened 2022-02-17
`/play end` should just be `/play`


rafaeldamasceno:
## Rationale

The `/play end` suncommand should still be named the same, since it performs exactly the same as the old `/play`. Having it as subcommand creates confusion, since it isn't obvious that `end` refers to the main play command with the default and expected behavior.

## Description

Rework the subcommand `/play end` into the old `/play`.

## Addendum

I'm not sure whether the existence of subcommands implies that the parent command cannot be called standalone. If that's the case, I'd very much suggest the current subcommands be moved to a different parent command and keeping the `/play` command as is.

afonsojramos:
Okay, I just proposed this to be the solution to #148, so I think we can kill two birds with one stone here as I agree that we should keep the normal `play` command as simply `/play`.

rafaeldamasceno:
> What name should we give to this advanced play command? How about `/splay` for "super play"?

That's a good suggestion (although in my mind it was suggested from "special play"). Another alternative is "advanced play", so `/aplay`, `/advplay` or any other variant.

danrpinho:
> What name should we give to this advanced play command? How about `/splay` for "super play"?

I feel that using `/splay` would be inconsistent with other commands such as, for example, `/forceskip`. I wouldn't mind if this advanced play command had a longer name (like `/superplay` or `/advancedplay`), since Discord has an autocomplete function and we could just use tab-complete to access it.

aquelemiguel:
@danrpinho I really like `/superplay`!

Issue #146, opened 2022-02-17
Allow disabling the DJ role


rafaeldamasceno:
## Rationale

When the DJ roles were introduced, no alternative for keeping the old behavior was implemented (besides manually assigning the role to all members of a guild).

## Description

Allow, through a configuration command, disabling the DJ role as a requirement to use restricted commands. This allows a guild to keep the old bot behavior, should they wish so. Additionally, if implemented, make it the default behavior (so that the role isn't created unnecessarily).

## Reference

[feat: custom DJ Role for parrot commands #115](https://github.com/aquelemiguel/parrot/pull/115)

PR #151, opened 2022-02-17
chore: split `play` responsibilities & require perms for destructive commands

### LINKED ISSUE #148: Destructive `/play` subcommands aren't restricted
## Description

Some newly added `/play` subcommands have destructive effects on the queue and should be restricted to DJ role members to keep in line with the other commands that do the same thing. They are `/play next` and `/play jump`.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The commands should be restricted to DJ role members. |
| **Observed** | The commands are unrestriced. |

## Repro Steps

1. Make sure you don't have the DJ role.
2. Verify you are able to use the `/play next` and `/play jump` subcommands.

## Environment

Key | Value
-- | --
Operating System | Ubuntu 21.10 impish
Kernel | x86_64 Linux 5.13.0-28-generic
CPU | Intel Core i5-6500 @ 4x 3,6GHz
RAM | 15893MiB

afonsojramos:
Closes #148 and #149.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -15,3 +15,20 @@ pub mod skip;
 pub mod stop;
 pub mod summon;
 pub mod version;
+
+#[derive(Clone, Copy)]
+pub enum PlayMode {
+    End,
+    Next,
+    All,
+    Reverse,
+    Shuffle,
+    Jump,
+}
+
+#[derive(Clone, Copy)]
+pub enum QueryType {
+    Keywords,
+    VideoLink,
+    PlaylistLink,
+}
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -15,20 +15,3 @@ pub mod skip;
 pub mod stop;
 pub mod summon;
 pub mod version;
-
-#[derive(Clone, Copy)]
-pub enum PlayMode {
-    End,
-    Next,
-    All,
-    Reverse,
-    Shuffle,
-    Jump,
-}
-
-#[derive(Clone, Copy)]
-pub enum QueryType {
-    Keywords,
-    VideoLink,
-    PlaylistLink,
-}
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -88,7 +88,7 @@ pub async fn play(
     drop(handler);
 
     match mode {
-        PlayMode::End => match query_type {
+        Mode::End => match query_type {
             QueryType::Keywords | QueryType::VideoLink => {
                 enqueue_song(ctx, &call, guild_id, url.to_string(), query_type).await;
             }
@@ -96,7 +96,7 @@ pub async fn play(
                 enqueue_playlist(ctx, &call, guild_id, url, mode, query_type).await;
             }
         },
-        PlayMode::Next => match query_type {
+        Mode::Next => match query_type {
             QueryType::Keywords | QueryType::VideoLink => {
                 enqueue_song(ctx, &call, guild_id, url.to_string(), query_type).await;
                 rotate_tracks(&call, 1).await;
@@ -109,7 +109,7 @@ pub async fn play(
                 }
             }
         },
-        PlayMode::Jump => match query_type {
+        Mode::Jump => match query_type {
             QueryType::Keywords | QueryType::VideoLink => {
                 enqueue_song(ctx, &call, guild_id, url.to_string(), query_type).await;
 
@@ -129,7 +129,7 @@ pub async fn play(
                 }
             }
         },
-        PlayMode::All | PlayMode::Reverse | PlayMode::Shuffle => match query_type {
+        Mode::All | Mode::Reverse | Mode::Shuffle => match query_type {
             QueryType::VideoLink | QueryType::PlaylistLink => {
                 enqueue_playlist(ctx, &call, guild_id, url, mode, query_type).await;
             }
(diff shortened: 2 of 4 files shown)


joao-conde (commented):
I dislike the `splay` command. I would:

- keep the same level of permissions for all play (sub)commands (either all DJ or not)
- kickstart discussion on Serenity to implement permissions at the subcommand level

joao-conde:
I agree they are destructive, hence I would just protect them all. Changing the queue is always dangerous.

afonsojramos:
But just plainly adding songs to the end of the queue seems like something anyone should be able to do... 🤔  I'm not 100% sure though 🤪 We must find some middle ground... Any suggestions @rafaeldamasceno?

joao-conde:
I dislike the splay due to naming, overall split from other play commands which do the same, etc...

I agree everyone should be able to do /play and /play jump should be restricted.

Hence, the best solution for me would actually be to REVERSE all these flag/subcommand shenanigan and just give a regular play (if you want to play jump just /play + /skip you lazy dark web users). But this is very drastic so I dont see us all agreeing on this.

rafaeldamasceno:
> I agree everyone should be able to do /play and /play jump should be restricted.

100% in agreement with you here. If an API limitation is why the subcommands can't be restricted then some other things need to be taken in consideration.

> I dislike the splay due to naming, overall split from other play commands which do the same, etc...

Well, I also don't love the naming. In fact, since slash commands have suggestions, I'd say we can be explicit in the naming: `/superplay` is a good option since super is a short word. I disagree however they do the same. Specially for `/play all`, it has a clearly different behavior that cannot be achieved by any other command.

>Hence, the best solution for me would actually be to REVERSE all these flag/subcommand shenanigan and just give a regular play (if you want to play jump just /play + /skip you lazy dark web users). But this is very drastic so I dont see us all agreeing on this.
I tend to agree: I never appreciated this subcommand approach, and I stated this clearly. But the fault in this case doesn't rely on the approach so much, but in the way the feature itself interacts with the DJ role and it's implementation.

In any case, I feel like this discussion shouldn't be held here. @afonsojramos @aquelemiguel Where could we discuss this whole feature interaction more in depth? Perhaps a discussion ticket?

afonsojramos:
> Hence, the best solution for me would actually be to REVERSE all these flag/subcommand shenanigan and just give a regular play (if you want to play jump just /play + /skip you lazy dark web users). But this is very drastic so I dont see us all agreeing on this.

The thing is that we will loose the functionality of some of the commands :( 

Another proposal would be to proceed with this recommendation from @joao-conde and revert all the subcommands, BUT, introduce a `skipto`/`forceskipto` in order to also provide the jump functionality. However, we would still lose the `/play all` which I think it is actually useful. But at the same time that can be achieved by requiring for the traditional playlist link.

afonsojramos:
Okay, I just had an idea that might actually be feasible... What if we manually checked for a user's role? We know the role's name, we just need to check it depending on which subcommand is called. This does not, however, fix the `/play` instead of `/play end` ordeal.

rafaeldamasceno:
> What if we manually checked for a user's role? We know the role's name, we just need to check it depending on which subcommand is called.

This is what I was going to suggest. I wanted to do this discussion further in depth because I understand that sometimes you (actual developers of the bot) are trying to align with the Discord API design values. Reworking the flag feature into subcommands or leaving all the play subcommands restricted just because it's what works when looking at the available API prevents a lot of useful functionality or good UX from existing. There are scenarios when you need to subvert the API/slash command design, because it simply wasn't designed to take into account every possible use case.

For this specific case, I would leave all `/play` or `/splay` or whatever subcommands unrestricted and manually check that the user has the necessary role for the `jump` and `next`, like @afonsojramos just said. Yes, it looks like any user can call them, but it's just two subcommands out of the bunch that do destructive actions, so it doesn't bother me that it looks like any user can call them, but instead face a permission error message. Restricting the commands that a user can definitely not use qualifies as "good enough" usage of the API for me. I wouldn't make the permissions in the bot limited to what the API allows in the command restriction department.

However, I really don't use the DJ role, not even in a big guild. If #146 is going to be closed and if the old unrestricted behavior is brought back as default, I think that people that actually use the DJ role more should jump in and give their feedback. Even so, I think that mine and Afonso's suggestion is good.

As for the `/play`/`/play end` situation, is it really not possible for `/play` to exist if there are subcommands? If it isn't, I don't see any way to solve this but to have the subcommands moved to another command.

joao-conde:
I had the same idea @afonsojramos . But manually checking for the role won't make the commands appear "grayed" out like the current commands for which a user does not have permission. It would instead just send a channel message like "no permission". 

But then we would have grayed out commands and other commands which you could call but then would be denied access, seems hella weird.

rafaeldamasceno:
Yeah, it's weird the behavior is inconsistent, but its the best we can do. As I've said in the previous comment, I think we should try to apply the API and slash command design as much as possible, but to use our own solutions when it doesn't solve our problems.

aquelemiguel:
@afonsojramos @joao-conde @rafaeldamasceno I've only now read this thread. Let me chime in for a sec.

Firstly, I have no idea why we're discussing this on a PR, this conversation should have been held on #148.

Secondly, obviously we're not reverting the subcommand implementation. If anything, migrate them to a special command like `/splay` as suggested above.

Lastly, arguably the most destructive command is the base play itself because anyone can enqueue a 100+ song playlist without any restrictions. So obviously, the entire `/play` subcommand group should be limited. You don't go to the club and play your Spotify playlist, that's the DJ's job.

I'm closing this PR because it's clearly not agreed on and to move extra discussion to the appropriate place.

PR #156, opened 2022-02-18
fix: revert audio format back to `webm[abr>0]/bestaudio/best`

aquelemiguel:
Switch back to default audio format as livestreaming wasn't working because of `-f bestaudio`.
(Thank you to @danrpinho for reporting this!)


afonsojramos (commented):
Looking good, live videos are working again 👌  Let me test a bit more tomorrow.

PR #158, opened 2022-02-22
feat: `skip-to` optional parameter 

afonsojramos:


src/handlers/serenity.rs
```diff
@@ -253,7 +249,15 @@ impl SerenityHandler {
                 })
                 .create_application_command(|command| {
                     command.name("skip").description("Skips the current track")
-                    .default_permission(true)
+                    .default_permission(false)
+                    .create_option(|option| {
+                        option
+                            .name("skip-to")
+                            .description("Track to skip to")
```
aquelemiguel, line 256: "Track **index** to skip to" is a little more explicative.

src/handlers/serenity.rs
```diff
@@ -253,7 +249,15 @@ impl SerenityHandler {
                 })
                 .create_application_command(|command| {
                     command.name("skip").description("Skips the current track")
```
aquelemiguel, line 251: If we have an optional argument for the track index, this description should probably read "Starts a vote to skip **to a specific track**" or something more broad.

src/commands/skip.rs
```diff
@@ -1,76 +1,39 @@
...
-    let channel_guild_users = guild_users
-        .into_values()
-        .filter(|v| v.channel_id.unwrap() == bot_channel_id);
-    let skip_threshold = channel_guild_users.count() / 2;
-
-    if cache.current_skip_votes.len() >= skip_threshold {
+    } else {
+        for _ in 1..to_skip {
+            queue.dequeue(1);
+        }
```
aquelemiguel, line 32: Dangerous approach because if the user uses an index that exceeds the size of the queue, it loops without reason. But worse, if you input something stupid like 100M the bot crashes (most likely because it takes the handler and doesn't release it after it's done looping).

I'd suggest changing `to_skip` to be `min(input, queue_size)` and then calling `modify_queue` because it's more readable (and should be more performant?).

aquelemiguel (changes requested):

src/commands/skip.rs
```diff
@@ -34,52 +26,14 @@ pub async fn skip(
...
-    let guild_users = ctx.cache.guild(guild_id).await.unwrap().voice_states;
-    let channel_guild_users = guild_users
-        .into_values()
-        .filter(|v| v.channel_id.unwrap() == bot_channel_id);
-    let skip_threshold = channel_guild_users.count() / 2;
-
-    if cache.current_skip_votes.len() >= skip_threshold {
+    } else {
         for _ in 1..tracks_to_skip {
             queue.dequeue(1);
```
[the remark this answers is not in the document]
afonsojramos: That's why it doesn't ;) The playing track is the index 0, I just remove n-1 tracks, and only then do I skip the currently playing one.

src/commands/skip.rs
```diff
@@ -34,52 +26,14 @@ pub async fn skip(
...
         }
         if queue.skip().is_ok() {
-            create_response(&ctx.http, interaction, SKIPPED).await?
+            return create_response(&ctx.http, interaction, SKIPPED).await;
```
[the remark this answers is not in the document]
afonsojramos: I wouldn't say so because of two things.
1. It would add a bit of unnecessary complexity.
2. If the skip-to index is bigger than the queue, track number X will not exist.
However, if you agree, I can maybe do a "Skipped to {Music Name}".

src/commands/voteskip.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub async fn forget_skip_votes(data: &Arc<RwLock<TypeMap>>, guild_id: GuildId) -> Result<(), ()> {
+    let mut data = data.write().await;
+
+    let cache_map = data.get_mut::<GuildCacheMap>().ok_or(())?;
+    let cache = cache_map.get_mut(&guild_id).ok_or(())?;
+    cache.current_skip_votes = HashSet::new();
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
afonsojramos: Well, I kind of can't because `skip.rs` still exists, but I'll take that into account next time!

src/handlers/serenity.rs
afonsojramos: You can find that in the `voteskip` command!

src/commands/skip.rs
afonsojramos: Nope, `modify_queue` is not more performant, and definitely not more readable in my opinion, since dequeue actually calls `modify_queue`.

Either way, I tested with 1000 and it handled it fine, but I'll implement said limiter.

aquelemiguel: > Nope, modify_queue is not more performant, and definitely not more readable in my opinion, since dequeue actually calls modify_queue.

The queue is actually a `VecDeque`, represented like `[tail]..unused..[head]` in memory. This means popping from the front is super quick. It's the recalculating of tail/head that was concerning me.

I was suggesting [draining](https://doc.rust-lang.org/src/alloc/collections/vec_deque/mod.rs.html#1249-1251) because it looks more optimized than repeatedly [removing](https://doc.rust-lang.org/src/alloc/collections/vec_deque/mod.rs.html#1854) (note Serenity calls this and not `pop_front` if the index is 1) and is undoubtedly more idiomatic.

However, I took the liberty of measuring both approaches but the differences are negligible and vary too much.

> Either way, I tested with 1000 and it handled it fine, but I'll implement said limiter.

Well yeah, 1000 is a breeze but 100M isn't.

src/commands/skip.rs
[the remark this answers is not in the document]
aquelemiguel: > If the skip-to index is bigger than the queue, track number X will not exist.

Take the minimum as you calculated before, no?

> However, if you agree, I can maybe do a "Skipped to {Music Name}".

I think I prefer this, yeah. But as you wish. 🙂

src/commands/skip.rs
joao-conde: I think anyways the `to_skip` needs to be minimum limited, so as not to waste loops dequeueing stuff that does not exist plus now allowing an user to crash the both if he wants to.

aquelemiguel: Didn't even notice you kept the skip afterwards. Given that you're removing from the middle, drain seems even more obvious now... Also why not incorporate the current track into that range (making sure you stop it beforehand)?

src/handlers/serenity.rs
aquelemiguel: I'm a little confused with the reply. I'm saying this description is sort of misleading because you can also skip up to a certain track, not just the currently playing.

image

`voteskip`'s description is correct IMO!

afonsojramos: > I'm a little confused with the reply. I'm saying this description is sort of misleading because you can also skip up to a certain track, not just the currently playing.

Sure, but there is no vote with the skip command 😅 Additionally, the base command does skip the currently playing track, so I think it makes sense.

src/commands/skip.rs
afonsojramos: > Also why not incorporate the current track into that range (making sure you stop it beforehand)?

This was because I wanted to make sure that the skip is almost unnoticeable, and removes the overhead that might be caused by the removal of the tracks in the queue before it.

src/commands/skip.rs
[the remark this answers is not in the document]
afonsojramos: Done ✅

aquelemiguel (approved):

PR #160, opened 2022-02-23
feat: remove range of tracks

### LINKED ISSUE #155: Add range end to `/remove`
## Rationale

Sometimes you want to remove an entire range of songs from the queue.

## Description

Add an optional positive integer range end parameter. When set, instead of removing a single track from the queue, remove all tracks in the given range.

afonsojramos:
Closes #155.


src/handlers/serenity.rs
```diff
@@ -217,6 +217,14 @@ impl SerenityHandler {
                                 .required(true)
                                 .min_int_value(1)
                         })
+                        .create_option(|option| {
+                            option
+                                .name("until")
+                                .description("Add a subsequent track position to remove a range of tracks")
```
aquelemiguel, line 223: We aren't describing options in declarative style, maybe something "**Upper range track position to remove multiple tracks**" is more suitable.

aquelemiguel (changes requested):
Base functionality is good, just needs a few tweaks and out-of-bounds checks.

src/handlers/serenity.rs
afonsojramos: I like it!

src/commands/remove.rs
```diff
@@ -30,21 +34,34 @@ pub async fn remove(
...
     } else if queue.len() < remove_index + 1 {
         create_response(&ctx.http, interaction, FAIL_NO_SONG_ON_INDEX).await
+    } else if remove_until < remove_index {
+        create_response(&ctx.http, interaction, FAIL_REMOVE_RANGE).await
     } else {
         let track = queue.get(remove_index).unwrap();
-        handler.queue().dequeue(remove_index);
 
+        handler.queue().modify_queue(|v| {
+            v.drain(remove_index..=min(remove_until, queue.len() - 1));
```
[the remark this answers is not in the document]
afonsojramos: Works and nice edge case! Pushing the fix!

aquelemiguel (approved):

PR #161, opened 2022-02-24
refactor: superplay logic

### LINKED ISSUE #159: `/queue` live updating should happen after queue manipulation
Queue isn't live updating correctly with some `/superplay` commands, because the updating is being done at `enqueue_song` level but songs switch position afterwards in some commands, like `/superplay next`.

aquelemiguel:
Closes #159.

* Introduces `insert_track` method for inserting a track in a specific position of the queue.
* Removes `enqueue_playlist` method.
* When working with playlists, immediately insert each track in the right position instead of rotating in the end.

### DIFF
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -14,10 +14,14 @@ use crate::{
 use serenity::{
     builder::CreateEmbed,
     client::Context,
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):

afonsojramos:
Short and to the point, I like it! 🥇

PR #162, opened 2022-02-25
fix: clippy for rust `1.59.0`

afonsojramos:
Currently the [CI's clippy is failing](https://github.com/aquelemiguel/parrot/runs/5327167853) due to the release of [Rust `1.59.0`](https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html). This PR fixes those errors.

- [x] Moved from **container specific** builds to **toolchain specific** builds to allow for a permanent build process with rust's `stable` toolchain. This also follows the build process that is also used in Serenity.
- [x] Started testing only for the **MRSV** (Minimum supported rust version, which was tested to be 1.54.0 due to some nightly features that were only available from this version forward) and **Stable**.
- [x] Applied `1.59.0` clippy fixes.
- [x] Bumped `version` to `1.3.0`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,18 @@ jobs:
     name: Build
     strategy:
       matrix:
-        rust-version: ["1.55", "1.56", "1.57", "1.58", "1.59"]
+        rust-version: ["1.54", "stable"]
     runs-on: ubuntu-latest
-    container: rust:${{ matrix.rust-version }}
     steps:
       - name: Repository Checkout
         uses: actions/checkout@v2
+
+      - name: Install Toolchain
+        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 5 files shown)


afonsojramos:
Ran some tests with several Rust versions and it seems that 1.54 is our MRSV (Minimum supported rust version), should we maybe move towards only testing the MRSV and Stable? @aquelemiguel

aquelemiguel:
> Ran some tests with several Rust versions and it seems that 1.54 is our MRSV (Minimum supported rust version), should we maybe move towards only testing the MRSV and Stable? @aquelemiguel

I see no problems with that, go ahead! 🚀

joao-conde (approved):

aquelemiguel (approved):

PR #166, opened 2022-03-05
feat: better error handling

joao-conde:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Create the crate error type: `ParrotError`. Replace all commands return type with results where the error type is `ParrotError`. This allows usage of the `?` operator. <br>Remove all error messages from the commands and centralize error message logic. Currently, each type of error implements the `Display` trait which allows it to be printed or formatted into a string. This string is sent as a simple message as the reply to the interaction. <br>Create also the verify function, which verifies if the first argument evaluates to valid/truth: `true`, `Some(_)` or `Ok(_)`. <br><br><br>**Food for thought:** In the future, we can use the same logic for command return types of `Ok(_)` and handle the interaction responses in a single place. |


src/commands/seek.rs
```diff
@@ -31,17 +31,16 @@ pub async fn seek(
             .and_then(|token| token.parse::<u64>().ok()),
     );
 
-    if minutes.is_none() || seconds.is_none() {
-        return create_response(&ctx.http, interaction, FAIL_TIMESTAMP_PARSING).await;
-    }
+    verify(minutes, ParrotError::Other(FAIL_TIMESTAMP_PARSING))?;
+    verify(seconds, ParrotError::Other(FAIL_TIMESTAMP_PARSING))?;
```
afonsojramos, line 35: Can't we do this check in the same line?
```suggestion
    verify(minutes && seconds, ParrotError::Other(FAIL_TIMESTAMP_PARSING))?;
```

src/errors.rs
```diff
@@ -0,0 +1,83 @@
...
+            ParrotError::WrongVoiceChannel => f.write_str(FAIL_WRONG_CHANNEL),
+            ParrotError::AuthorNotFound => f.write_str(FAIL_AUTHOR_NOT_FOUND),
+            ParrotError::AlreadyConnected(mention) => {
+                f.write_fmt(format_args!("{} {}", FAIL_ANOTHER_CHANNEL, mention))
+            }
+            ParrotError::NothingPlaying => f.write_str(NOTHING_IS_PLAYING),
+            ParrotError::Serenity(err) => f.write_str(&format!("{err}")),
+        }
+    }
+}
```
afonsojramos, line 47: Love this!! 💯

afonsojramos (approved):
Loved this 💯 The readability of our code is going to improve by a lot with this I think!

src/commands/seek.rs
joao-conde: No. That does not compile.

joao-conde: They are both options, each can be converted to a bool, but not implicitly. What could be done is 

```rust
verify(minutes.to_bool() && seconds.to_bool(), ParrotError::Other(FAIL_TIMESTAMP_PARSING))?;
```

afonsojramos: But does that not fail when one of them is not defined? I would prefer that over two different checks for the same error.

aquelemiguel: If anything, this is a little better no? I have no strong opinions on either.

```rust
verify(minutes.is_none() && seconds.is_none(), ParrotError::Other(FAIL_TIMESTAMP_PARSING))?;
```

afonsojramos: Well, yes that one is better, but then we need to negate the verification.

aquelemiguel (approved):

src/commands/seek.rs
aquelemiguel: Ah, true. Edited.

joao-conde: Sure, just defeats the purpose of implementing ToBool trait for options. I might be able to cook something up where instead of receiving a ToBool it receives an expression where elements implement ToBool. I will try it later tonight

afonsojramos: But there are multiple occurrences where you need to validate more than one thing for the same error, which is why I asked!

joao-conde: Took the opportunity to have more specific messages.

PR #176, opened 2022-03-15
fix: idle disconnect based on top track status

### LINKED ISSUE #171: Parrot is not disconnecting for inactivity
## Description

After a whole night, the bot stayed in the room without disconnecting.

image

This happened because of the following: 
image

Note: When I first called the `/queue` it was still playing, except it was paused. After I resumed, within a second it stopped the music and somehow skipped the song.

## Expected vs. Observed

| - | - |
| --- | --- |
| **Expected** | The bot disconnects after a while with no one in the room. |
| **Observed** | The bot stayed through a whole night, all alone. |

aquelemiguel:
Closes #171. 

## Decisions
The old logic was bad, it was only counting up if the queue was empty. This means that if a track was paused, it'd never disconnect (because the queue would never clear). Now it takes into account the status of the top track.

### DIFF
--- a/src/handlers/idle.rs
+++ b/src/handlers/idle.rs
@@ -2,7 +2,7 @@ use serenity::{
     async_trait, http::Http,
     model::interactions::application_command::ApplicationCommandInteraction,
 };
-use songbird::{Event, EventContext, EventHandler, Songbird};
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):
Thanks! Looking good 💯

PR #177, opened 2022-03-17
docs: add dependency badge

afonsojramos:
Add the following badge to the readme. We are currently MAYBE INSECURE :eyes:

[dependency status](https://deps.rs/repo/github/aquelemiguel/parrot)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@
   A hassle-free, highly performant, host-it-yourself Discord music bot
 </p>
 
-[![dependency status](https://deps.rs/repo/github/aquelemiguel/parrot/status.svg)](https://deps.rs/repo/github/aquelemiguel/parrot)
+<p align="center">
+  <a href="https://deps.rs/repo/github/aquelemiguel/parrot"><img src="https://deps.rs/repo/github/aquelemiguel/parrot/status.svg"></a>
+</p>
 
 ## Deployment
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-  A hassle-free, highly performant, host-it-yourself Discord music bot  
+  A hassle-free, highly performant, host-it-yourself Discord music bot
 </p>
 
 [![dependency status](https://deps.rs/repo/github/aquelemiguel/parrot/status.svg)](https://deps.rs/repo/github/aquelemiguel/parrot)
--- a/README.md
+++ b/README.md
@@ -3,8 +3,10 @@
 </p>
 
 <p align="center">
-  A hassle-free, highly performant, host-it-yourself Discord music bot
+  A hassle-free, highly performant, host-it-yourself Discord music bot  
 </p>
+
+[![dependency status](https://deps.rs/repo/github/aquelemiguel/parrot/status.svg)](https://deps.rs/repo/github/aquelemiguel/parrot)
 
 ## Deployment
 


aquelemiguel:
Sorry for the late review.

I appreciate the effort, but I don't really like the positioning of this badge. Maybe center it?
Could also be a good chance to add a CI badge?

joao-conde (commented):
I like it, but it may be also a good time to fix these dependency warnings. Maybe we can bump the tokio and regex packages?

aquelemiguel (commented):
Sorry for the late review.

I appreciate the effort, but I don't really like the positioning of this badge. Maybe center it?
Could also be a good chance to add a CI badge?

afonsojramos:
Regarding the dependencies, I don't think we should mix it. This PR is purely related to the badge and that's it.

Regarding the aesthetic, I will make some tries and put the attempts here.

joao-conde:
Does this read from the Cargo TOML or Lock? If it is reading the TOML we shouldn't have this. 

For example, for the regex crate, we specify "1.5" meaning we want the latest patch. If we rebuild and update the cargo lock like in https://github.com/aquelemiguel/parrot/pull/179 we successfully get the latest regex patch ("1.5.5") which fixes the security vulnerability. The badge should check the lock because of this. And updating the Cargo TOML to restrict to "1.5.5" is not that good.

joao-conde:
It seems I have answered myself. Interesting enough, take a look at the master version of `tokio`:

https://github.com/aquelemiguel/parrot/blob/main/Cargo.lock#L1785
https://github.com/aquelemiguel/parrot/blob/main/Cargo.toml#L29

I don't know how this happened, but seems like we already use the recent one in our lock and therefore I would assume whenever we run parrot. So on https://github.com/aquelemiguel/parrot/pull/179 I have updated the TOML and we should have a secure repository after.

afonsojramos:
It does check the TOML and it is why in our case we have a "MAYBE Insecure". If you check other repos you will see that only when you have a fixed version that is insecure does it truly show up as insecure. 

https://deps.rs/repo/github/denoland/deno

joao-conde:
> It does check the TOML and it is why in our case we have a "MAYBE Insecure". If you check other repos you will see that only when you have a fixed version that is insecure does it truly show up as insecure.
> 
> https://deps.rs/repo/github/denoland/deno

Then I don't think we can have this. For example, right now it complains regex is insecure. That is incorrect because in TOML we have 1.5 which means it fetches the latest patch, which is 1.5.5 so it should not be marked as insecure.

afonsojramos:
We should however, lock patch versions in `cargo.toml`, which is why I just opened #180. Unless you can make sure that all the libraries that we depend on follow semver and that they won't introduce new features in a patch (spoiler alert, you can't - and we've actually done it more than once on our package because you don't want to bump the minor), then you should always lock the versions that you are using.

joao-conde:
I will just say it again if people can't respect SEMVER and introduce breaking changes in patch versions, they are evil. But by all means, if it is something that happens, let's lock it.

afonsojramos:
Yes, but in that case we are evil. Check [v1.2.1](https://github.com/aquelemiguel/parrot/releases/tag/v1.2.1).

joao-conde:
> Yes, but in that case we are evil. Check [v1.2.1](https://github.com/aquelemiguel/parrot/releases/tag/v1.2.1).

I clearly did not agree with such numbering then

joao-conde (approved):

joao-conde:
Anyways, this badge is cool. Try some more @afonsojramos maybe like for CI and stuff.

afonsojramos:
@joao-conde stop doing merges on other peoples PRs 😠

joao-conde:
> @joao-conde stop doing merges on other peoples PRs 😠

https://github.com/aquelemiguel/parrot/pull/166

https://github.com/aquelemiguel/parrot/pull/166/commits/15d540cbf2e1f079857c4d4d37fa6a0f23684993

aquelemiguel (approved):

afonsojramos:
I know my mistakes in the past, what matters is the choices you take in the future 👾

afonsojramos:
Also, for CI I believe that this is enough nowadays

image

aquelemiguel:
I missed these email spams, we should work on Parrot more often.

joao-conde:
I didn't consider it a mistake, I just did it as you did it. But we can stop that, sure. Agreed only now :)

PR #184, opened 2022-03-31
fix: unexpected idle disconnects

### LINKED ISSUE #181: Unexpected idle disconnects
### 📝 Description

While its logic is correct, #176 introduced a bug where the bot unexpectedly idle disconnects. This is because sometimes skipping a track does not remove it from the queue, but instead keeps it at the top in a paused state. This makes the idle counter not reset and thus may prompt the bot to disconnect earlier than expected if this bug persists on the following tracks.

We should investigate how Songbird handles skipping a track. We've had this issue in the past, which forced us to implement the `force_skip_top_track` method.

### 🪜 Reproduction Steps

1. Trace the counter in `idle.rs`
2. Queue multiple songs
3. Arbitrarily skip and notice that sometimes the counter keeps incrementing

- **Parrot version**: `v1.3.1`
- **Operating System**: `Windows`

### 📸 Screenshots

_No response_

aquelemiguel:
Closes #181.

### DIFF
--- a/src/handlers/idle.rs
+++ b/src/handlers/idle.rs
@@ -22,12 +22,16 @@ pub struct IdleHandler {
 impl EventHandler for IdleHandler {
     async fn act(&self, ctx: &EventContext<'_>) -> Option<Event> {
         if let EventContext::Track(track_list) = ctx {
-            if let Some(top_track) = track_list.first() {
-                // if the top track is playing (not paused), then reset the counter
-                if matches!(top_track.0.playing, PlayMode::Play) {
-                    self.count.store(0, Ordering::Relaxed);
-                    return None;
-                }
+            // looks like the track list isn't ordered here, so the first track in the list isn't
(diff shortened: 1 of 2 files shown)


src/handlers/idle.rs
```diff
@@ -22,12 +22,16 @@ pub struct IdleHandler {
...
-                // if the top track is playing (not paused), then reset the counter
-                if matches!(top_track.0.playing, PlayMode::Play) {
-                    self.count.store(0, Ordering::Relaxed);
-                    return None;
-                }
+            // looks like the track list isn't ordered here, so the first track in the list isn't
+            // guaranteed to be the first track in the actual queue, so search the entire list
+            let bot_is_playing = track_list
+                .iter()
+                .any(|track| matches!(track.0.playing, PlayMode::Play));
```
joao-conde, line 29: Smart boy

joao-conde (approved):

Issue #123, opened 2022-02-01
Message Translations


joao-conde:
## Rationale

Add translations to change the default parrot language from English to something else.

## Description

- Each message sent to the guild server must be translated. 
- It should default to English.
- New translations can be easily added by open source contributors

joao-conde:
@aquelemiguel @afonsojramos I think we discussed this at a certain point. I have an idea of how to do it. Leave it to me if you agree with the issue.

afonsojramos:
Yeah, the big move to simply using the `strings.rs` file was mostly to address this. Go ahead!

aquelemiguel:
Sure, removing other contributors from this.

joao-conde:
> Yeah, the big move to mostly using the `strings.rs` file was mostly to help with this. Go ahead!

Yes, but at the same time, the translations will no longer be stored in code and therefore not part of the release binary. This would make no sense for tons of translations. 

Currently thinking of storing the translations in JSON files making it easier to change or add new ones without changing the codebase.

afonsojramos:
I actually had already investigated a bit on how to do this properly with community integrations and stuff and I've found that platforms such as https://crowdin.com/ or https://weblate.org/en-gb/ offer a really good interface for managing the translations. Maybe it is a bit overkill, what do you think?

joao-conde:
Seems a bit overkill to me but nonetheless when I have time I will look into them.

joao-conde:
Alright, my proposed solution is one that:
1 - does not increase source and binary sizes with each new translation/locale
2 - can be easily contributed by open source devs
3 - custom translations can be easily be made locally

With that in mind, I propose the following:
- a `messaging.rs` module which implements a key `message(message_type, locale = "en")` function
- in that module, we keep the base strings we use right now as discord messages (they will be the default, English)
- the idea is `message_type` will be an enum with the types of messages we send (the queue is empty, the user is not in the channel, etc.)
- `locale` is the typical language ISO code; if none is given, use the default constants we have in `strings.rs` right now
- the translations should be stored in JSON files that are read on startup into a hashmap and used for translating from message_type + locale -> string/message

Obviously, this means the binary of parrot alone when run will always default to English. The language translations are only available if the deployed parrot instance has local JSON translation files. Alternatively, we can serve the JSON files somewhere and make HTTP requests to fetch them, but I find this overkill. 

What do you think @afonsojramos  @aquelemiguel

aquelemiguel:
[...]

PR #198, opened 2022-05-29
refactor: parrot message types

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/aquelemiguel/parrot/issues/123 |
| Dependencies | -- |
| Decisions | In the world of rust and typed languages, we should make more usage of types. Sending ad-hoc strings as messages (mythical `strings.rs`) worked fine but we can do better. It becomes a bigger problem now due to https://github.com/aquelemiguel/parrot/issues/123 where we want to have mappings of translations. This means if I want to send the HelloWorld message I can do so in fr, en, or pt. If all I use is a string, I have no way of knowing which message it is and how to translate it. This PR preps the translations PR. <br> Here we introduce the concept of message types. These types are then displayed. A future message translation implementation will take into account the server's set language and return a different string to display. <br> Another important topic is error handling. I did it in a way where every error was bubbled up. While intended at the time, I see now its a bit weird for the user to receive messages like "invalid token" or something of the sorts. A next PR will make the automatic conversion from ParrotError -> ParrotMessage -> sent to user. This way, we can filter specific error messages we want to display and convert those into a specific ParrotMessage and for the other errors, we just throw a generic error message. |


afonsojramos (approved):
Looking good!! Way cleaner now! Quality PR 💯

PR #202, opened 2022-10-15
fix: make errors follow up messages

### LINKED ISSUE #201: Interaction has already been acknowledged.
### 📝 Description

So there's been a change to the way interaction responses are handled which causes some issues with the error reporting. 
If we attempt to create a new response and there's already an existing response the following is returned:

```
parrot  | thread 'tokio-runtime-worker' panicked at 'failed to create response: Serenity(Http(UnsuccessfulRequest(ErrorResponse { status_code: 400, url: Url { [TRUNCATED] }, error: DiscordJsonError { code: 40060, message: "Interaction has already been acknowledged.", errors: [] } })))', src/handlers/serenity.rs:373:14
```

It's an easy fix, but I figured I would ask how we'd like to address it:
A) Clobber the old response with the new one
B) Create a followup message when creating the initial response fails

Unfortunately I don't know of a clean way to implement this with the current error handling built into the create_embed_response at the moment. 

### 🪜 Reproduction Steps

1. Play a video that has been privated or has been deleted (`https://www.youtube.com/watch?v=cwFQpRTwaP0` for example
2. The application will respond indicating that it is searching
3. yt-dlp will throw an error since the video is inaccessible
4. The application will attempt to respond with an error but will fail because the initial response still exists

- **Parrot version**: `main`
- **Operating System**: `Arch Linux`

### 📸 Screenshots

_No response_

StaticRocket:
Make errors follow up messages instead of responses so they do not collide with any other responses.

Resolves #201

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -33,6 +33,16 @@ pub async fn create_response_text(
     create_embed_response(http, interaction, embed).await
 }
 
+pub async fn create_followup_text(
+    http: &Arc<Http>,
+    interaction: &mut ApplicationCommandInteraction,
+    content: &str,
+) -> Result<Message, ParrotError> {
+    let mut embed = CreateEmbed::default();
+    embed.description(content);
+    create_embed_followup(http, interaction, embed).await
+}
+
 pub async fn edit_response(
     http: &Arc<Http>,
     interaction: &mut ApplicationCommandInteraction,
@@ -64,6 +74,17 @@ pub async fn create_embed_response(
                 .kind(InteractionResponseType::ChannelMessageWithSource)
                 .interaction_response_data(|message| message.add_embed(embed))
         })
+        .await
+        .map_err(Into::into)
+}
+
+pub async fn create_embed_followup(
+    http: &Arc<Http>,
+    interaction: &mut ApplicationCommandInteraction,
+    embed: CreateEmbed,
+) -> Result<Message, ParrotError> {
+    interaction
+        .create_followup_message(&http, |followup| followup.add_embed(embed))
         .await
         .map_err(Into::into)
 }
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -33,6 +33,20 @@ pub async fn create_response_text(
     create_embed_response(http, interaction, embed).await
 }
 
+pub async fn create_error_text(
```
joao-conde, line 36: Can we just call it "create_followup_message" as this can be used for other uses in the future and not only errors

src/utils.rs
```diff
@@ -33,6 +33,20 @@ pub async fn create_response_text(
...
+    http: &Arc<Http>,
+    interaction: &mut ApplicationCommandInteraction,
+    content: &str,
+) -> Result<(), ParrotError> {
+    let mut embed = CreateEmbed::default();
+    embed.description(content);
+    interaction
+        .create_followup_message(&http, |followup| followup.add_embed(embed))
+        .await
+        .map(|_| ())
```
joao-conde, line 46: Do we need this mapping? Can we terminate with .await?

joao-conde (changes requested):

src/utils.rs
StaticRocket: It was required to be a drop in replacement for create_response_text, but I'll drop it to make this more generic for future follow up message usage

src/utils.rs
StaticRocket: Sure, do you want me go ahead and implement the equivalent functions for create_response, edit_response, edit_response_text, edit_response_text, and edit_embed_response as well or just what we need for now?

joao-conde: Just what we need for now and yeah I agree

joao-conde (approved):

joao-conde:
Thanks @StaticRocket , on point as always

PR #208, opened 2022-10-25
fix: `/superplay next` with an empty queue throws ParrotError::NotInRange

### LINKED ISSUE #207: `/superplay next` with an empty queue throws ParrotError::NotInRange
### 📝 Description

The `insert_track` function should account for the queue possibly being empty. Right now, it's responding with a confusing (and incorrect) error. In that case, it should just insert the track on the first position.

### 🪜 Reproduction Steps

1. Ensure the queue is empty.
2. Call **/superplay next query: asdf**.
3. Parrot should reply with "_index should be between 1 and 0 but was 1_", which is incorrect. Instead, it should just insert the track to the first position.

- **Parrot version**: `v1.4.2`
- **Operating System**: Windows 11

### 📸 Screenshots

image

aquelemiguel:
Resolves #207. Also removed the NotInRange check altogether as there's no real way it can fail.

### DIFF
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -365,10 +365,10 @@ async fn insert_track(
     let queue_size = handler.queue().len();
     drop(handler);
 
-    verify(
-        idx != 0 && idx < queue_size,
-        ParrotError::NotInRange("index", idx as isize, 1, queue_size as isize),
-    )?;
+    if queue_size <= 1 {
+        let queue = enqueue_track(call, query_type).await?;
+        return Ok(queue);
+    }
 
     enqueue_track(call, query_type).await?;
 
--- a/src/commands/play.rs
+++ b/src/commands/play.rs
@@ -370,6 +370,11 @@ async fn insert_track(
         return Ok(queue);
     }
 
+    verify(
+        idx > 0 && idx <= queue_size,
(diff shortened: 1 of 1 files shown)


src/commands/play.rs
```diff
@@ -365,10 +365,10 @@ async fn insert_track(
     let queue_size = handler.queue().len();
     drop(handler);
 
-    verify(
-        idx != 0 && idx < queue_size,
-        ParrotError::NotInRange("index", idx as isize, 1, queue_size as isize),
-    )?;
```
joao-conde, line 371: Yeah but insert 5 on a queue of length 3 now is not checked right?

aquelemiguel: Correct, but looking at these function calls I can't really see how that scenario is possible. Am I overlooking something?

joao-conde: If insert is called with our index command?

aquelemiguel: You mean a command for inserting a track in a specific position in the queue? We don't have that.

aquelemiguel: Added the assertion back. Idx is always >=1, so having the assert before the new code would fail (1 <= 0). So I added it as a safeguard for the already existing one.

joao-conde (approved):

afonsojramos (approved):

PR #223, opened 2022-11-14
feat: allow streaming from other yt-dlp supported sources

### LINKED ISSUE #209: Allow streaming from other yt-dlp supported sources
### 📝 Description

yt-dlp allows for streaming from multiple other sources other than YouTube. [A lot more actually](https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md).

In the play command, we check the query/url in this fashion:

```rust
let query_type = if url.contains("spotify.com") {
  // extract metadata from the spotify link and search it on youtube
}
else if url.contains("youtube.com") {
  // pass the link directly to yt-dlp
}
else {
  // pass the provided query to yt-dlp
}
```

There are a few issues with this approach:
* It limits the source to YouTube, meaning a user would not be able to stream from SoundCloud, Twitter or any other media providing website supported by yt-dlp.
* It does not take into account shortened links that redirect to YouTube (like bit.ly or more flagrant, youtu.be) and searches the link in YouTube, which is pretty dumb.
* It does not include URL validation. "/play query: ftp://youtube.com" isn't a valid link to pass to yt-dlp, for instance. We should ensure the link is valid before passing it to yt-dlp.

### 🪜 Reproduction Steps

N/A

- **Parrot version**: `v1.4.2`
- **Operating System**: Windows 11

### 📸 Screenshots

_No response_

aquelemiguel:
Closes #209.

### What's changed

Introduces a new command `/managesources` to allow or ban domains. Below is what it looks like:

image

This is what invoking `/play` on a banned domain looks like:

image

The `/np` command changed to accommodate the source of the playback:

image


src/commands/play.rs
```diff
@@ -77,16 +79,60 @@ pub async fn play(
...
+                Some(Spotify::extract(spotify, url).await?)
+            }
+            Some(other) => {
+                let mut data = ctx.data.write().await;
+                let settings = data.get_mut::<GuildSettingsMap>().unwrap();
+                let guild_settings = settings.entry(guild_id).or_default();
+
+                let other = other.to_string().replace("www.", "");
+
+                if guild_settings.banned_domains.contains(&other)
```
StaticRocket, line 97: Domain parsing will have to be a little more complex than this unfortunately. Here the domain `play.dev` being in the blacklist will mean that `play.dev.com` also gets blacklisted, but these are two different domains. See the following for my best take at a domain parsing algorithm in rust:
```rs
use std::iter;
use url::Url;

fn compare_domains(one: &str, two: &str) -> bool {
    let iterator_one = one.rsplit_terminator(".");
    let iterator_two = two.rsplit_terminator(".");
    let mut matching = true;
    for (element_one, element_two) in iter::zip(iterator_one, iterator_two) {
        println!("{} == {}", element_one, element_two);
        matching = matching && element_one == element_two ;
    }
    matching
}

fn main() {
    let test = Url::parse("http://www.youtube.com./").unwrap();
    let test1 = Url::parse("http://youtube.com/").unwrap();
    println!("{}", compare_domains(test1.domain().unwrap(), test.domain().unwrap()));
}
```

That rsplit_terminator is necessary because `google.com.` is just as valid as `google.com` and will be resolved the same.

StaticRocket: Wait, I'm dumb. You are using a HashSet for this. That bypasses my issue with the caveat that the user will have to individually add all necessary subdomains to a block/allow list.

StaticRocket (approved):
Looks good to me. I left my comment, but it's probably not worth acknowledging. If we actually wanted to play into hierarchical domain block/allow lists properly then it'll need to prioritize more specific entries and that is a bit more functionality than what is needed here.

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> Result<(), ParrotError> {
+    let guild_id = interaction.guild_id.unwrap();
+
+    let mut data = ctx.data.write().await;
+    let settings = data.get_mut::<GuildSettingsMap>().unwrap();
+
+    // fetch the allowed domains and transform to a comma-separated string
+    let guild_settings = settings.entry(guild_id).or_default();
+
+    // transform the domain sets from the settings into a string
```
joao-conde, line 35: comments miss aligned

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+            guild_settings.update_domains();
+
+            // it's now safe to close the modal, so send a response to it
+            int.create_interaction_response(&ctx.http, |r| {
+                r.kind(InteractionResponseType::DeferredUpdateMessage)
+            })
+            .await
+            .ok();
+        })
+        .collect::<Vec<_>>() // streams do nothing unless polled
```
joao-conde, line 129: superfluous comment

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+        .label(DOMAIN_FORM_BANNED_TITLE)
+        .custom_id("banned_domains")
+        .style(InputTextStyle::Paragraph)
+        .placeholder(DOMAIN_FORM_BANNED_PLACEHOLDER)
+        .value(banned_str)
+        .required(false);
+
+    let mut components = CreateComponents::default();
+
+    components
```
joao-conde, line 74: perhaps `default().create.create...`  or remove newline

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+        .label(DOMAIN_FORM_ALLOWED_TITLE)
+        .custom_id("allowed_domains")
+        .style(InputTextStyle::Paragraph)
+        .placeholder(DOMAIN_FORM_ALLOWED_PLACEHOLDER)
+        .value(allowed_str)
+        .required(false);
+
+    let mut banned_input = CreateInputText::default();
+
+    banned_input
```
joao-conde, line 64: perhaps `default().create.create...`  or remove newline

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+        .clone()
+        .into_iter()
+        .collect::<Vec<String>>()
+        .join(";");
+
+    drop(data);
+
+    let mut allowed_input = CreateInputText::default();
+
+    allowed_input
```
joao-conde, line 54: perhaps `default().create.create...`  or remove newline

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+                d.title(DOMAIN_FORM_TITLE);
+                d.custom_id("manage_domains");
+                d.set_components(components)
+            })
+        })
+        .await?;
+
+    // collect the submitted data...
+    let collector = ModalInteractionCollectorBuilder::new(ctx)
+        .filter(|int| int.data.custom_id == "manage_domains") // only keep submissions from the domain modal
```
joao-conde, line 91: superfluous comment because the filter seems clear enough

src/commands/manage_sources.rs
```diff
@@ -0,0 +1,133 @@
...
+            r.kind(InteractionResponseType::Modal);
+            r.interaction_response_data(|d| {
+                d.title(DOMAIN_FORM_TITLE);
+                d.custom_id("manage_domains");
+                d.set_components(components)
+            })
+        })
+        .await?;
+
+    // collect the submitted data...
```
joao-conde, line 89: remove ...

joao-conde (changes requested):

src/commands/play.rs
StaticRocket: >  google.com. is just as valid as google.com and will be resolved the same

This issue, however, is still present.

afonsojramos (approved):

src/commands/play.rs
aquelemiguel: Thank you for the review, @StaticRocket. 

I think you're correct, regardless! I do use a HashSet but blocking `play.dev` won't block `play.dev.com`, it has to strictly match. The set just prevents duplicate entries on the form.

Regarding your parsing algorithm, I like it! However, one of the URLs (the one inputted by the user) cannot be parsed via `Url::parse` because it's missing the `http://` portion - panicking with:

```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RelativeUrlWithoutBase', src/main.rs:18:43
```

My question is, should the user manually add all subdomains or should it be done automatically? Also, should the user inputted domains be checked for validity?

StaticRocket: > Regarding your parsing algorithm, I like it! However, one of the URLs (the one inputted by the user) cannot be parsed via Url::parse because it's missing the http:// portion

That's why it's written to compare string versions of the domain (we shouldn't have to parse the user input as we are using lazy matching). We also allow the user to wildcard match all domains by using the domain entry `.` with that system.

> My question is, should the user manually add all subdomains or should it be done automatically? Also, should the user inputted domains be checked for validity?

I believe subdomain's should inherit parent domain status unless explicitly stated otherwise. The "explicitly stated otherwise" part is the only thing my algorithm can't handle because it would have to rank entries based on specificity (g.z.t.example.com should be interpreted after example.com and override example.com if matched).

StaticRocket: I wonder if there's any nice domain parsing logic we could lift from ftl / pihole ...

aquelemiguel: > That's why it's written to compare string versions of the domain ; ) (we shouldn't have to parse the user input as we are using lazy matching). We also allow the user to wildcard match all domains by using the domain entry . with that system.

I see what you mean, I was misinterpreting your code, sorry!

Piggybacking on your suggestion, I've been testing the PR with the check below, `domain` being the blocked domain and `subdomain` being the request:

```rust
pub fn compare_domains(domain: &str, subdomain: &str) -> bool {
    subdomain == domain || subdomain.ends_with(domain)
}
```

For instance, if `m.youtube.com` is blocked but `youtube.com` is requested, it's allowed. Conversingly, if `youtube.com` is blocked but `m.youtube.com` is requested, it's blocked. If both happen to be blocked, it's not conflicting, even if redundant.

I reckon it should be enough for now? Did you have any other edge case in mind?

src/guild/settings.rs
```diff
@@ -1,10 +1,60 @@
...
     pub autopause: bool,
+    pub allowed_domains: HashSet<String>,
+    pub banned_domains: HashSet<String>,
+}
+
+pub trait Update {
+    fn set_allowed_domains(&mut self, allowed_str: &str);
+    fn set_banned_domains(&mut self, banned_str: &str);
+    fn update_domains(&mut self);
+}
```
joao-conde, line 15: Why are we making this a trait instead of GuildSettings associated functions? Not reusable.

src/guild/settings.rs
```diff
@@ -1,10 +1,60 @@
...
+
+impl Default for GuildSettings {
+    fn default() -> GuildSettings {
+        GuildSettings {
+            autopause: false,
+            allowed_domains: HashSet::from(["youtube.com".to_string()]),
```
joao-conde, line 21: Put the default allowed domain in a const top level variable for visibility and easy  change

src/guild/settings.rs
```diff
@@ -1,10 +1,60 @@
...
+            banned_domains: HashSet::new(),
+        }
+    }
+}
+
+impl Update for GuildSettings {
+    fn set_allowed_domains(&mut self, allowed_str: &str) {
+        let allowed: HashSet<String> = allowed_str
```
joao-conde, line 29: Type can probably be inferred ?

src/guild/settings.rs
```diff
@@ -1,10 +1,60 @@
...
+            .split(';')
+            .filter(|s| !s.is_empty())
+            .map(|s| s.to_string())
+            .collect();
+
+        self.allowed_domains = allowed;
+    }
+
+    fn set_banned_domains(&mut self, banned_str: &str) {
+        let banned: HashSet<String> = banned_str
```
joao-conde, line 39: Type can probably be inferred ?

joao-conde (changes requested):

joao-conde (approved):

src/handlers/serenity.rs
```diff
@@ -78,11 +78,6 @@ impl SerenityHandler {
     async fn create_commands(&self, ctx: &Context) -> Vec<Command> {
         Command::set_global_application_commands(&ctx.http, |commands| {
             commands
-                .create_application_command(|command| {
-                    command
-                        .name("autopause")
-                        .description("Toggles whether to pause after a song ends")
-                })
```
joao-conde, line 85: Why are we deleting autopause?!?!?

aquelemiguel: ClownClownMemeGIF

Issue #227, opened 2022-12-23
The bot does not compile


rafaeldamasceno:
### 📝 Description

Title is pretty self explanatory :) 

### 🪜 Reproduction Steps

1. Try to run the CI

### ℹ Environment / Computer Info

GitHub Actions

### 📸 Screenshots

https://github.com/rafaeldamasceno/parrot/actions/runs/3765059132/jobs/6400242903

afonsojramos:
This has to do with the new Rust version most likely. But I am checking the logs on my mobile so I might be wrong.

blimp4242:
I'm experiencing the same issue, errors are also the same. Did you find a workaround?

aquelemiguel:
Hi @rafaeldamasceno @blimp4242, I've looked into this and found it's related with [rspotify](https://crates.io/crates/rspotify).
I'm pushing a fix in a bit, bumping it to 0.11.6 and fixing the code accordingly, which should make the CI pass.

I am however really confused about how our CI is automatically assuming we're on the latest version of this crate...
Could you have a look at what's going on when you have time, @afonsojramos?

aquelemiguel:
@blimp4242 @rafaeldamasceno Thank you for the patience and inputs. [v1.5.1](https://github.com/aquelemiguel/parrot/releases/tag/v1.5.1) is now out, which fixes this.

PR #232, opened 2022-12-28
ci: keep Cargo.lock

rafaeldamasceno:
Fixes the actual reason that caused #227, which was having the dependencies lock in the `.dockerignore`. Also adds some small QoL changes like forcing cargo to use the locked dependencies, fixing the Rust Docker image version and adding a compose file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,11 @@ WORKDIR "/parrot"
 RUN mkdir src
 RUN echo "fn main() {}" > src/main.rs
 COPY Cargo.toml ./
-RUN cargo build --release
+COPY Cargo.lock ./
+RUN cargo build --release --locked
 
 COPY . .
-RUN cargo build --release
+RUN cargo build --release --locked
 
 # Release image
 # Necessary dependencies to run Parrot
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,4 +12,3 @@ README.md
 # Cargo files
 **/*.rs.bk
 /target/
-Cargo.lock


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build image
 # Necessary dependencies to build Parrot
-FROM rust:slim-bullseye as build
+FROM rust:1.65-slim-bullseye as build
```
afonsojramos, line 3:
```suggestion
FROM rust:slim-bullseye as build
```
Let's keep this as it is. This is kept like this to "force" us to be on rust's latest version.

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: "3"
+
+services:
+  parrot:
+    container_name: parrot
+    build: .
+    # image: ghcr.io/aquelemiguel/parrot
+    env_file: .env
+    restart: unless-stopped
```
afonsojramos, line 9: I'd delete this whole file. We do not need this.

aquelemiguel:
Changed title to conform with conventional commits. Passing this to @afonsojramos, he's better versed in this CI stuff.

afonsojramos (changes requested):
Looking good, just trim it down to the essentials

Dockerfile
rafaeldamasceno: Sure, but then what's the difference between this and the Build workflow? Should we force the build workflow to use the latest version as well?

afonsojramos: That's exactly why we also have the following:
```yaml
     matrix:
        rust-version: ['1.65', 'stable']
```

docker-compose.yml
rafaeldamasceno: Not sure I agree--nowadays, it's commonplace to include a compose file for single image projects. Running `docker-compose up` is pretty different from copying a really big `docker run` command from the documentation. It will also allow us to specifiy standard volume paths when implementing serialization for guild settings.

Dockerfile
rafaeldamasceno: Sorry, I'm not too familiar with the actions syntax. Does this mean it tries to use 1.65 then stable? Why not have stable only?

joao-conde: AFAIK we we're using the minimal version we support plus the latest stable one. Therefore, Parrot should work from 1.65 onwards to stable (or at least in 1.65 and stable, with 1.65 being the minimal, nothing below that is supported)

docker-compose.yml
joao-conde: Im with @afonsojramos. And even if it is handy in the future, we will cross that bridge when we get there. For the purpose of this PR, keep it to copying the cargo lock file

joao-conde (approved):

afonsojramos (approved):

PR #243, opened 2023-06-20
Enqueuing playlists: skip songs that cannot be added

jheuel:
This pull request changes the behaviour when enqueuing playlists. If there is a problem while enqueuing a track of a playlist the problematic song is skipped instead of canceling the whole process.
Also the mask for playlists in the YouTube::extract function was changed to match more playlist links.


joao-conde:
Thanks for the contribution @jheuel 

I think it makes perfect sense to try to enqueue the other songs.

joao-conde (approved):

PR #251, opened 2024-07-12
Add youtu.be as a default domain

kmgb:
| - | - |
| --- | --- |
| Issue | Somewhat related to https://github.com/aquelemiguel/parrot/issues/209 |
| Dependencies | none |
| Decisions | Users in my server found it confusing that youtu.be didn't work immediately. Since youtu.be is a first-party shortened URL provided by the "share" button on YouTube, it might make sense to just allow it by default. |


joao-conde (approved):
Thanks for the change!

